PR #1, opened 2017-12-27
Add Operator factory

mbartling:
Abstracted away the giant switch case hell. Code is about the same size, but now it should be easier to add an optimizer.

### DIFF
--- a/utensor_cgen/core.py
+++ b/utensor_cgen/core.py
@@ -10,6 +10,7 @@ from .snippets import register_template
 from .composer import Composer
 from ._snippets_base import SnippetContainer, Snippet
 from ._types import TF_TYPES_MAP
+from .operators import OperatorFactory
 
 __all__ = ["CodeGenerator"]
 
@@ -40,6 +41,8 @@ class CodeGenerator(object):
     container.template_vars["placeholders"] = []
     container.add_header('"{}"'.format(header_fname))
 
+    opFactory = OperatorFactory()
+
     print("Parsing {}".format(self.pb_file))
     graph_info, layers = parse_pb(self.pb_file)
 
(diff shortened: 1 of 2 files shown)


dboyliao:
Looking good.
I'll test it before I merge it.
I'll let you know I'm done ASAP 👍

utensor_cgen/core.py
```diff
@@ -40,6 +41,8 @@ def generate(self, src_fname: str):
     container.template_vars["placeholders"] = []
     container.add_header('"{}"'.format(header_fname))
 
+    opFactory = OperatorFactory()
+
     print("Parsing {}".format(self.pb_file))
     graph_info, layers = parse_pb(self.pb_file)
```
dboyliao: Why don't you refactor these switch cases, Placeholder and Const, altogether with you `OperatorFactory`?

utensor_cgen/operators.py
```diff
@@ -0,0 +1,125 @@
+from .snippets import *  # pylint: disable=W0401
+from .snippets import register_template
+from ._snippets_base import SnippetContainer, Snippet
```
dboyliao: I think this import is irrelevant here.

utensor_cgen/operators.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class _Operator():
+  # Base Class
+  snippet = None
+  name = ""
+
+  def getSnippet(self):
+    return self.snippet
```
dboyliao: Are you trying to make `self.snippet` read only so you write a getter here?
Why not use `@property`?

utensor_cgen/core.py
mbartling: Short answer, I didn't have much time to work on this. They depend on more than just `op_info` so I just added `**kwargs` to hand this in the future.

utensor_cgen/operators.py
mbartling: all 3? Don't I need the first?

utensor_cgen/operators.py
mbartling: not to familiar with `@property`, XD If it works better than we can change it :)

mbartling: yeah, `property` seems like the way to go.

utensor_cgen/core.py
mbartling: Before I move `Placeholder`, this logic needs to change 
```
container.template_vars["placeholders"].append(out_tname)
header_snippet.template_vars["placeholders"].append(out_tname)
```

Alternatively, I can just replace snippet with the `out_tname`.

dboyliao: I see.
I think it's reasonable to do it for now.
OK, I'll merge it after I test it all.
I haven't come up with a good way to automatic test it with a develop board, so I have to do it manually.
It'll take some time

mbartling: Yeah auto testing these things is a nightmare. Ideally this code shouldn't be mbed specific so it should just run in an arbitrary program.

PR #2, opened 2018-01-01
adding reference counting feature to code generation

neil-tan:
Proposing the generation of reference counting code.
Compiles but not tested.
graph-dividing eval() not yet implemented

Do not merge

### DIFF
--- a/utensor_cgen/core.py
+++ b/utensor_cgen/core.py
@@ -10,6 +10,7 @@ from .composer import Composer
 from ._snippets_base import SnippetContainer, Snippet
 from ._types import TF_TYPES_MAP
 from .operators import OperatorFactory
+from .optimizer import *
 
 __all__ = ["CodeGenerator"]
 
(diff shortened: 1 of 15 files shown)


utensor_cgen/optimizer.py
```diff
@@ -0,0 +1,15 @@
+from collections import Counter
+
+def get_refc_table(graph_info) -> Counter:
```
dboyliao: Type annotation is python 3 only.
This will break python2 support.

dboyliao: Type annotation is python 3 only.
This will break python2 support.

utensor_cgen/snippets.py
```diff
@@ -73,51 +73,56 @@ def _permute_args(args: list, perm: list=None):
 
 
 class AddOpSnippet(Snippet):
-  def __init__(self, inputs, output, tf_dtype=tf.float32):
+  def __init__(self, inputs, output, init_count, tf_dtype=tf.float32):
```
dboyliao: I think we should set `init_count` a default value, say, 0.
And let the optimizer to decide the desired `init_count`.

utensor_cgen/core.py
```diff
@@ -46,6 +47,12 @@ def generate(self, src_fname: str):
     print("Parsing {}".format(self.pb_file))
     graph_info, layers = parse_pb(self.pb_file)
 
+    ref_table = get_refc_table(graph_info)
```
dboyliao: Maybe leave the ref counting to a optimizer is a better design?
I'd like to talk to you about the interface of an optimizer.

PR #4, opened 2018-01-14
Feature/ref cnt

dboyliao:
Reference counting for memory efficiency


mbartling:
the final generation the constant output directory is not consistent.
For example:
```
void get_simple_graph_ctx(Context& ctx) {
{
    TensorIdxImporter t_import;
    ctx.add(t_import.float_import("model/y_0.idx"),
            "y:0",
            1);
}
```

This specifies `model/*` instead of `idx_data/*`

utensor_cgen/_pbparser_impl.py
```diff
@@ -89,26 +90,30 @@ def _parse_graph_nodes(graph_def):
...
-          tensor = graph.get_tensor_by_name(out_tensor)
-          op_info["output_content"][tensor.name] = tensor.eval()
+        for tensor_name, _, _ in output_tensor:
+          tensor = graph.get_tensor_by_name(tensor_name)
+          output_content[tensor_name] = tensor.eval()
+      graph_info[node.name] = OperationInfo(input_tensor, output_tensor, op_type, output_content)
   return graph_info
 
 
 def _parse_graph_layers(graph_def):
```
mbartling: Is this necessary? Can't we just trust the output names from the CLI input

utensor_cgen/_pbparser_impl.py
```diff
@@ -149,9 +154,35 @@ def _is_freeze_graph(graph_def):
   return is_frozen
 
 
-def _parse_graph_def(graph_def):
+def _parse_graph_nodes_bfs(graph_def, output_nodes=None):
+  if output_nodes is None:
```
mbartling: Should output nodes be required?

utensor_cgen/operators.py
```diff
@@ -13,107 +13,115 @@ def snippet(self):
 
 
 class _AddOperator(_Operator):
-  def __init__(self, op_info, **kwargs):
+  def __init__(self, op_info, ref_counts, to_eval):
```
mbartling: You could have left this as `**kwargs` as it will expand to `ref_counts` and `to_eval`

utensor_cgen/optimizer.py
```diff
@@ -0,0 +1,40 @@
+#-*- coding:utf8 -*-
+from collections import defaultdict
+
+
+class Optimizer(object):
+
+  @classmethod
+  def optimize(cls, ops_info, bfs_order, output_nodes, level='simple'):
```
mbartling: Need a default `level=None` for no optimization at all. Pretty sure this is handled by simple

utensor_cgen/optimizer.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    optimized_order = []
+
+    refcnt_table = cls._tensor_ref_count(ops_info)
+    for op_name in bfs_order[::-1]:
+      op_info = ops_info[op_name]
+      if op_name in output_nodes or op_info.op_type in ["Const", "Placeholder"]:
+        to_eval = False
+      else:
+        to_eval = True
```
mbartling: So evaluate after every operator?

utensor_cgen/template_env.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding:utf8 -*-
```
mbartling: What exactly does this file do, can you add comments

utensor_cgen/templates/snippets/argmax_op.cpp
```diff
@@ -1,6 +1,19 @@
+{% if create_sptr %}
```
mbartling: There is a lot of code reuse in these blocks. Let's make them another template and import it here

utensor_cgen/operators.py
dboyliao: I know.
But then I have to do something in the function body like this:
```
ref_counts = kwargs.get('ref_counts')
if ref_counts is None:
  raise ValueError(.....)
```

If you set them as positional arguments, it's required and checked by python interpreter.
I just want to make things clear

utensor_cgen/optimizer.py
dboyliao: Yes.
And let the reference counting do the job.

utensor_cgen/template_env.py
dboyliao: Oh, I just figure out that it's better to manage the template files with `Environment` object in jinja2.
http://jinja.pocoo.org/docs/2.10/api/

With it, you have better control over templates rendering.

utensor_cgen/templates/snippets/argmax_op.cpp
dboyliao: Yay, you are right.
I just didn't have time to figure out how to organize where to put the import template files.
So I just copy&paste for now.

Pipfile.lock
```diff
@@ -0,0 +1,245 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "126e63133d9f44caa3ae8c84ef58f56722c1b08fbbeddcbbf08730875ffe89b2"
+        },
+        "host-environment-markers": {
+            "implementation_name": "cpython",
+            "implementation_version": "3.6.3",
```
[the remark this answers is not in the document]
dboyliao: I think it's auto generated by `Pipfile`
I didn't specify any python version in the `Pipfile` so I don't think it's an issue.

requirements.txt
```diff
@@ -1,5 +0,0 @@
-Jinja2==2.10
```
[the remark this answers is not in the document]
dboyliao: Well, I prefer `pipenv`.
Besides, we have `pip` support.

utensor_cgen/__main__.py
```diff
@@ -25,17 +44,22 @@ def _build_parser():
...
                       metavar="EMBED_DIR", default=None,
                       help="the data dir on the develop board (default: the value as the value of -d/data-dir flag)")
+  parser.add_argument("--output-nodes", dest="output_nodes",
+                      type=_nargs(), metavar="node,node,...",
+                      default=None,
+                      help="list of output nodes")
   parser.add_argument("--debug-comment", dest="debug_cmt",
                       action="store_true",
                       help="Add debug comments in the output source file (default: %(default)s)")
+  parser.add_argument("-V", "--version", action="store_true", dest="version", help="show version")
```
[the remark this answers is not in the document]
dboyliao: Well, `v` can also mean `verbose`.
But if you insist, I don't mind to change it to `v`

utensor_cgen/_pbparser_impl.py
dboyliao: I think it would be nice if the cli can figure out the output nodes for me or I can overwrite it if necessary.
That's why I make it optional here.

utensor_cgen/_pbparser_impl.py
dboyliao: See my comment below.

utensor_cgen/__main__.py
```diff
@@ -1,20 +1,39 @@
...
 import os
-from .core import CodeGenerator
 
+from pip.commands.show import search_packages_info
+
+
+def main(pb_file, src_fname, idx_dir, embed_data_dir, debug_cmt, output_nodes, version):
+  if version:
+    pkg_version = next(search_packages_info(['utensor_cgen']))["version"]
+    print("\033[33mutensor_cgen version: {}\033[0m".format(pkg_version))
```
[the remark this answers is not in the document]
dboyliao: Maybe.
But it's just single line color print, I don't think it's a good time refactoring this part of code. XD
BTW, I'm planing for using `Click` to rewrite the cli in the future when the time is right.
http://click.pocoo.org/5/

requirements.txt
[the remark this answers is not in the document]
dboyliao: You mean we both keep `Pipfile` and a `requirements.txt`?

Pipfile.lock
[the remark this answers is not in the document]
dboyliao: Besides, I've test the code against both python 2 and 3.
`pipenv` is just one way to manage the dependency.
If users don't like it, they can choose `pip` or `setup.py`.

My point is: it's optional to users to choose how to install this program, so we don't have to make every options identical.

utensor_cgen/__main__.py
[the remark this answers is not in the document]
dboyliao: Just out of curious, which standard are you talking about?
Is there a standard for command line interface?

utensor_cgen/optimizer.py
dboyliao: You mean `level=None` for no optimization (no ref counting, ...etc)
And maybe `level='simple'` means simple ref counting?

Pipfile.lock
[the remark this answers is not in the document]
mbartling: cool i'm fine with that then. I just worry that devs may accidentally introduce python3 only code.

requirements.txt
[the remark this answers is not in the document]
mbartling: yes

utensor_cgen/__main__.py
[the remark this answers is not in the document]
mbartling: Ohh i like this

utensor_cgen/__main__.py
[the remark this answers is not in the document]
mbartling: There are common practices for linux cli. For example, the `docopt` package can generate cli from documentation.

utensor_cgen/_pbparser_impl.py
mbartling: cool, I just worry about improperly frozen graphs having the training cost functions etc.

utensor_cgen/operators.py
mbartling: Pretty sure there is python built ins for doing this cleanly. I think `**kwargs` is more manageable in the long run

utensor_cgen/optimizer.py
mbartling: yes

utensor_cgen/optimizer.py
mbartling: cool, I talked this over with @neil-tan it makes sense

utensor_cgen/template_env.py
mbartling: 👍

utensor_cgen/templates/snippets/argmax_op.cpp
mbartling: Cool I'll take a look at this

requirements.txt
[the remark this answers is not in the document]
dboyliao: Sure. I'll add `requirements.txt` later.
I'm working on other things.

utensor_cgen/__main__.py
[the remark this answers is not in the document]
dboyliao: Ok, so I'll leave the code as is for now.
In future commits, I'll rewrite the cli with `Click`

utensor_cgen/__main__.py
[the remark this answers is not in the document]
dboyliao: Alright, we'll go for `-v`.

utensor_cgen/operators.py
dboyliao: What is it exactly?
I can't agree with you based on your memory only.

utensor_cgen/optimizer.py
dboyliao: Ok, working on it

requirements.txt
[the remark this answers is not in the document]
mbartling: thanks

utensor_cgen/__main__.py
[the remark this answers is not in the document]
mbartling: works for me

utensor_cgen/__main__.py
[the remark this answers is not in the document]
mbartling: THE LITTLE THINGS :D

utensor_cgen/operators.py
mbartling: Well If you just do `kwargs["ref_counts"]` and reference counts is not in the dictionary then it will throw a `KeyError`

utensor_cgen/optimizer.py
mbartling: cool, I think it's ok to just note this for now and add it as a feature request in the issues tab.

utensor_cgen/operators.py
dboyliao: But a `KeyError` does not reflect the semantics here.
It's a **required** for the function to work properly.

Just think about the error message:
```
....
KeyError: 'ref_counts'
```
against
```
....
TypeError: __ini__ missing 1 required positional argument: 'ref_counts'
```

To me, the later is much much better than the former.

mbartling:
FYI, I think this is ready to merge

utensor_cgen/_pbparser_impl.py
dboyliao: Well, it's users' responsibility if they do not provide me the list of output nodes.

PR #8, opened 2018-01-17
Change bfs to topologic sort

mbartling:
BFS does not guarantee proper dependency resolution "@mbartling is a dumbass" -@mbartling

### DIFF
--- a/utensor_cgen/_pbparser_impl.py
+++ b/utensor_cgen/_pbparser_impl.py
@@ -154,7 +154,8 @@ def _is_freeze_graph(graph_def):
   return is_frozen
 
 
-def _parse_graph_nodes_bfs(graph_def, output_nodes=None):
+def _parse_graph_topologic_order(graph_def, output_nodes=None):
+  # https://en.wikipedia.org/wiki/Topological_sorting
   if output_nodes is None:
     output_nodes = _parse_graph_layers(graph_def)[-1]
   graph = Graph()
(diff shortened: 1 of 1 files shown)


mbartling:
FYI, neil and I verified that the code builds and produces correct result with:

```cpp
#include <mbed.h>
#include "FATFileSystem.h"
#include "SDBlockDevice.h"
#include "models/deep_mlp.hpp"
#include "tensorIdxImporter.hpp"

Serial pc(USBTX, USBRX, 115200);
SDBlockDevice bd(MBED_CONF_APP_SD_MOSI, MBED_CONF_APP_SD_MISO,
                 MBED_CONF_APP_SD_CLK, MBED_CONF_APP_SD_CS);
FATFileSystem fs("fs");

int main(void){
    bd.init();
    fs.mount(&bd);
    init_env(); // Needs to be moved to a model class

    Context ctx;

    //Tensor* myInput = TensorConstant({28,28}, 0);
    TensorIdxImporter t_importer;
    Tensor* myInput = t_importer.float_import("/fs/testData/deep_mlp/import-Placeholder_0.idx");
    get_deep_mlp_ctx(ctx, myInput);


    ctx.eval();
    S_TENSOR myOutput = ctx.get({"y_pred:0"});
    printf("Number guessed %d\n\r", *(myOutput->read<int>(0,0)));

}
```

dboyliao:
Just out of curious, why bfs doesn't guarantee proper dependency resolution?
Could you kindly provide an example?

mbartling:
@dboy when we were running `deep_mlp` we found that some inputs were not being resolved correctly. Also there is a formal reason why BFS doesn't work, but I can't remember what it is :P

dboyliao:
I see.
Since you guys test the code already, I'll merge it first.

mbartling:
💯

PR #11, opened 2018-02-15
Updated README

edoffagne:
Hello!
I have done a minor update in the README.
The pipenv command with # which suggests root while it should be run a normal user ($).
Thanks,
Erik


dboyliao:
Thanks.

PR #13, opened 2018-02-21
Add quantized add operator for getting started tutorial

mbartling:
This is needed for the tutorial. Quantized add can be added like this temporarily and fully implemented later. 

Works with utensor-mnist-demo

### DIFF
--- a/utensor_cgen/operators.py
+++ b/utensor_cgen/operators.py
@@ -131,6 +131,7 @@ class OperatorFactory():
                 "Dequantize": _DequantizeOperator,
                 "Max": _MaxOperator,
                 "Min": _MinOperator,
+                "QuantizedAdd", _AddOperator,
                 "QuantizeV2": _QuantizeV2Operator,
                 "QuantizedMatMul": _QuantizedMatMulOperator,
                 "QuantizedRelu": _QuantizedReluOperator,


utensor_cgen/operators.py
```diff
@@ -131,6 +131,7 @@ class OperatorFactory():
                 "Dequantize": _DequantizeOperator,
                 "Max": _MaxOperator,
                 "Min": _MinOperator,
+                "QuantizedAdd", _AddOperator,
```
dboyliao: It's a syntax error.
Should be
```
...
"QuantizedAdd": _AddOperator,
...
```

dboyliao: In addition to this syntax error, I don't think this hot fix will work anyway.
Since the QuantizedAdd takes 6 inputs but normal AddOp takes only 2 inputs.
Replacing QuantizedAdd with AddOp won't work anyway.

Here is the screen shot of the code generated with your PR, it just push too much inputs to the AddOp in `uTensor`:
screen shot 2018-02-24 at 2 45 16 pm

The definition of AddOp in `uTensor`:
https://github.com/uTensor/uTensor/blob/master/uTensor/ops/MathOps.hpp#L376

dboyliao: Before QuantizedAdd is available in `uTensor`, I'd like to recommend users to use a script on `TensorFlow` GitHub repo:
https://raw.githubusercontent.com/tensorflow/tensorflow/master/tensorflow/tools/quantization/quantize_graph.py

This script will not add QuantizedAdd to your graph as `TransformGraph` does.
I know it's not the standard way to do it, but it works fine for now.

dboyliao: And this project is aimed to provide tool which can generate a "working" cpp/hpp implementation in `uTensor` for users.
What I mean by "working" is compilable and runnable on a MCU.
As a result, I prefer not to merge the PR.

PR #24, opened 2018-04-12
B/remove eval

mbartling:
Got rid of the evil eval that was causing problems all over with feed_dict.

### DIFF
--- a/utensor_cgen/parser/_pbparser_impl.py
+++ b/utensor_cgen/parser/_pbparser_impl.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 
 import numpy as np
 from tensorflow import Graph, Session, import_graph_def
+from tensorflow.python.framework import tensor_util
 
 
 def _parse_tensor_name(tname):
@@ -107,9 +108,7 @@ def _parse_graph_info(graph_def):
       output_content = {}
       op_attr = node.attr
       if node.op in ["Const"]:
-        for tensor_name, _, _ in output_tensor:
-          tensor = graph.get_tensor_by_name(tensor_name)
-          output_content[tensor_name] = tensor.eval()
+        output_content[tensor_name] = tensor_util.MakeNdarray(node.attr['value'].tensor)
       graph_info[node.name] = OperationInfo(input_tensor,
                                             output_tensor,
                                             op_type,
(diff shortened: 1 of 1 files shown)


utensor_cgen/parser/_pbparser_impl.py
```diff
@@ -109,7 +110,8 @@ def _parse_graph_info(graph_def):
       if node.op in ["Const"]:
         for tensor_name, _, _ in output_tensor:
           tensor = graph.get_tensor_by_name(tensor_name)
-          output_content[tensor_name] = tensor.eval()
+          #output_content[tensor_name] = tensor.eval()
```
dboyliao: Just delete this line of code.
Good job.

utensor_cgen/parser/_pbparser_impl.py
```diff
@@ -109,7 +110,8 @@ def _parse_graph_info(graph_def):
       if node.op in ["Const"]:
         for tensor_name, _, _ in output_tensor:
           tensor = graph.get_tensor_by_name(tensor_name)
```
dboyliao: And remove `tensor = graph.get_tensor_by_name(...)` too.
It's not needed anymore.

mbartling:
@dboyliao fixed

utensor_cgen/parser/_pbparser_impl.py
```diff
@@ -107,9 +108,7 @@ def _parse_graph_info(graph_def):
       output_content = {}
       op_attr = node.attr
       if node.op in ["Const"]:
-        for tensor_name, _, _ in output_tensor:
-          tensor = graph.get_tensor_by_name(tensor_name)
-          output_content[tensor_name] = tensor.eval()
+        output_content[tensor_name] = tensor_util.MakeNdarray(node.attr['value'].tensor)
```
dboyliao: You goes too far XD
You still need to loop over `output_tensor`.
Now the `tensor_name` is undefined. LOL

dboyliao:
BTW, I think you should use `tensorflow.contrib.util.make_ndarray` instead of `MakeNdarray`.
Though they are just aliases now but `tensorflow.python.framework` is considered private.
(https://github.com/tensorflow/tensorflow/issues/798)
And they are not publicly well documented which means Google team might want to modify it in the future (I hope not....)
On the other hand, `tensorflow.contrib.util` is well(?) documented:
https://www.tensorflow.org/api_docs/python/tf/contrib/util

PR #30, opened 2018-05-12
Update tests

dboyliao:


utensor_cgen/code_generator.py
```diff
@@ -6,7 +6,7 @@
 import tensorflow as tf
 
 from .operators import OperatorFactory
-from .optimizer import Optimizer
+from .parser.optimizer import Optimizer
```
neil-tan: Why move this into parser?

utensor_cgen/parser/pbparser.py
```diff
@@ -44,8 +44,8 @@ def parse_pb(file_or_path, output_nodes=None):
   graph_def.ParseFromString(fid.read())
   fid.close()
 
-  if output_nodes is not None:
-    graph_def = graph_util.extract_sub_graph(graph_def, output_nodes)
+  sub_graph_def = graph_util.extract_sub_graph(graph_def,
```
neil-tan: What's this doing?

PR #37, opened 2018-07-09
Develop

dboyliao:
`docker` support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM ubuntu
+
+ENV UTENSOR_CGEN_BRANCH=develop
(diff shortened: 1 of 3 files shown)


mbartling:
FYI I have a dockerfile for building utensor as well and another for mbed builds, we should talk about adding a docker compose setup

neil-tan:
Assigning this to @mbartling as he is more familiar with the setup.

dboyliao:
@mbartling 
For now I use this dockefile for the final backup plan for the user if all other setup methods fail.
I'm not familiar with docker compose but I think we can work on it later.

BTW, I make a PR which shrink the docker image you use for circle-ci in `uTensor`.
Just a few simple fixes.

PR #40, opened 2018-07-20
Click cli

dboyliao:
cli implement with click library


mbartling:
I'm guessing it's because click is easier to use?

dboyliao:
@mbartling 
You can have a look here
https://github.com/uTensor/utensor_cgen/pull/40/files#diff-8972ab8db4e4a06ec324eefe647ead29

As you can see, it's much more easier to implement subcommand with `click`.
In addition, each subcommand can be used as normal python function.
It'll be easier to write unit tests for the cli than ordinary `argparse` style.

BTW, I also update the cnn demo in this PR for Coscup, a big developer conference in Taiwan.

PR #46, opened 2018-09-28
Added extra node properties to 'utensor-cli show' output: operation and output

mlubinsky-arm:
Added extra node properties to 'utensor-cli show' output: operation and output


neil-tan:
@dboyliao   Yo, we have a new contributor who is interested in taking on the `utensor-cli show` function. Could you please have a look, as a starting point?

PR #47, opened 2018-10-08
Add GraphViz dot visualizations for IR uGraphs.

mbartling:
Add the ability to FINALLY visualize graphs in uTensor code gen. Defaults to writing graphs to ./graphs/graph.gv.pdf

`brew install graphviz`
`myGraph.viz_graph()`
graph

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,6 +6,7 @@ name = "pypi"
 [packages]
 "e1839a8" = {path = ".", editable = true}
 setuptools = "==39.1.0"
+graphviz = "*"
 
 [dev-packages]
 pylint = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "57133a0a80cf07bdf986a881a181b545ec17a46afefb158e8d0046fd2e1839bb"
+            "sha256": "cf08b718cc12e4b14842f918027a90586850d3b4732f3f33a516dbdc400b9a0a"
         },
         "pipfile-spec": 6,
         "requires": {},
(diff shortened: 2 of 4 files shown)


neil-tan:
@mbartling awesome stuff. Is there a path to extend this to display op_type, dimension, etc?

mbartling:
@neil-tan It should be trivial to make these extensions, Graphviz is pretty powerful. I just made this quick PR as a placeholder for future extensions etc. I think just the general graph and node names is sufficient to help with like 95% of our bugs.

dboyliao:
Looks good.
Maybe I'll patch this PR so the tensor names go to the edges and node names go to the nodes in the graph image.

Now it's hard to infer what do we mean by a "graph".
Can be either a computational graph or literally an image. XD

dboyliao:
One quick thought: I don't think it should be a method of `uTensorGraph`.
It should be a function or a method of other objects (such as a `uTensorGraphVisualizer`, I assume).

I feel like it's against the single responsibility principle which means we're making `uTensorGraph` being responsible for too many tasks.
In your implementation, `uTensorGraph` need to carry computation informations (`ops`, `tensors`, blah blah blah) and the visualization.
That's too much.

mbartling:
@dboyliao I am not convinced that this should be a separate module or method as it is solely for debugging purposes. Every graph we have has a different interface (tf graphs, torch, etc.) and the one golden unifier is our IR graph. Therefore, it makes sense to keep this here.

dboyliao:
If it's solely for debugging purpose, why not add to a debugging util module?
As you adding more and more irrelevant methods (other than generating code) to the graph, it takes more time for future developer to read it.

dboyliao:
Take `tensorflow` as example, they didn't make visualization tool as part of TF itself but rather separate it out.
Not even a python submodule but another python package, `tensorboard`.
If following your argument, why not they make the visualization/analysis tools all methods of `tf.Graph`?
And you can imagine, how complex `tf.Graph` would be if they did.

mbartling:
I am ok with adding it to a debugging module, currently it is treated as debugging as the import only occurs on the function call. I just want to make sure this tool is available to our devs. Neil and I found it extremely valuable when doing the CMSIS stuff

dboyliao:
Since the viz method is in the ir code already, closing this PR.

PR #49, opened 2018-11-13
Pytorch

dboyliao:
`pytorch` support via `onnx`


mbartling:
@dboyliao The code changes make sense, but can you add a brief write up to your PR discussing:

- What high level changes needed to be made to support pytorch
- How pytorch graphs differ from tensorflow
- General plans/guide for adding more frontend parsers.

dboyliao:
@mbartling 
Right, I should elaborate it more.
Will do this night.

dboyliao:
```
+------------+     +------------------+       ===================
| model file | --> | front end Parser | --> || uTensorGraph (IR) || 
+------------+     +------------------+       ===================
                                                      |
               +-------------------------------+      |
               | graph optimizer (transformer) | <---‘ 
               +-------------------------------+
                                |
                                v
                  ==========================
                || uTensorGraph (optimized) ||
                  ==========================
                                   |
+--------------------------+       |
| backend (code generator) | <----‘  
+--------------------------+
     |
     `---> (target files, ex: model.cpp, model.hpp, weights.idx)
```
It's the overall design of `utensor_cgen`.
In this PR, I finish refactoring frontend parser.
So `utensor_cgen` can support various model files from different frameworks / libraries by adding different frontend parsers.

To be specific, the developer who want to extend frontend parser has to do
1. Implement a parser class
    - It should be a subclass of `utensor_cgen.frontend.base.Parser`
    - overwrite the abstract class method `parse(cls, fname: str, output_nodes: List[str])`. This class method is responsible taking a file name and a list of output node names and returning a instance of `uTensorGraph`
    - put the `.py` file of the frontend parser under `utensor_cgen/frontend` submodule. All `.py` files in this submodule will be imported automatically to validate the registration.
2. Register the parser class
    - with `register(target_exts: List[str])` decorator of `utensor_cgen.frontend.FrontendSelector`
    - `target_exts` is a list of string which are the target file extensions of model files. For example, it should be, by convention, `['.pb', '.pbtxt']` for `Tensorflow`.

At the time of this PR being made, the `backend` part is not cleanly refactored.
I'm working on this part and trying to make it as easy as possible to support different target libraries (other then `uTensor`) or languages (I'm really sick of C++).

cc @mbartling @neil-tan @Knight-X

dboyliao:
As for the difference of `pytorch` and `tensorflow`, the former is dynamic graph and the later is static graph.
That is, graph in `pytorch` is defined by the nodes and its metadata (you can track back one node's parent nodes with it) and graph in `tensorflow` is defined by one graph object (`tensorflow.Graph`).

In this PR, `pytorch` support is provided by `onnx` which by itself is a static graph equivalent to `pytorch.Module`. You can convert a `Module` in `pytorch` to `onnx` format natively (builtin in `pytorch`).

The best practice here is to write parser for `onnx` but now I simply leverage `onnx-tf` which is a python package parsing `onnx` files as `tensorflow.GraphDef`.
Then I use my `tensorflow.GraphDefParser` to parse the resulting `GraphDef` object.
It's not the best solution, however it's the fastest way I can think of to make it work.
I'll come back to this later when I know more about how `pytorch` manage `Module` objects internally.

dboyliao:
BTW, there is a internal api in `pytorch.onnx` called `_model_to_graph` which convert a `Module` to a graph (a `pytorch` graph ....?)

I can use it rather than `onnx-tf` but I just don't feel right using internal api.
That's why I go for `onnx-tf` and use my tensorflow parser.

Knight-X:
I think that changes look good. However, should we have a layer to legalize the graph before transformer pass? It will replace the unsupported node to the corresponding one.

dboyliao:
I was thinking it just another transformer (converting uTensorGraph to another uTensorGraph).
Maybe I should rename the optimization block.

dboyliao:
@Knight-X how is it now?

utensor_cgen/cli.py
```diff
@@ -104,13 +104,9 @@ def show_graph(model_file, **kwargs):
 
 def _show_pb_file(pb_file, **kwargs):
```
mbartling: Cant torch graphs also be represented in protobuf formats? This naming might get confusing

utensor_cgen/frontend/tensorflow.py
```diff
@@ -0,0 +1,86 @@
...
+import numpy as np
+from google.protobuf import text_format
+
+from utensor_cgen.frontend.base import Parser
+from utensor_cgen.frontend import FrontendSelector
+from utensor_cgen.ir.base import TensorInfo, OperationInfo, uTensorGraph
+from utensor_cgen.utils import topologic_order_graph
+
+
+@FrontendSelector.register(target_exts=['.pb', '.pbtxt'])
```
mbartling: I'm not a fan of selecting the frontend parsers based on file extensions. If anything, we need to load from protobuf and determine what kind of data structure, ugraph, custom_graph, etc. is described by the file and proceed from there.

dboyliao: So do you have any suggestion how to determine which frontend to use other than file extension?

utensor_cgen/cli.py
dboyliao: Which naming?
`_show_pb_file`?

FYI, yes, a torch graph can be represented in protobuf format.
First you save torch graph in `onnx` format and convert `onnx` to `GraphDef` via `onnx-tf`.

utensor_cgen/frontend/tensorflow.py
dboyliao: As far as what I know, there is no common way to detect the content format.
Something like: if the first few bytes is blah then TF, if blah then onnx else blah blah blah.
So file extension is the simplest way I can think of.

utensor_cgen/cli.py
mbartling: Cool, let's go ahead and leave it as is, but make it a TODO in the issue tracker.

utensor_cgen/frontend/tensorflow.py
mbartling: May be just reading the first "node" in the pb file. If it is a graph_def it should just be a list of NodeDefs. As long as the first entry is a NodeDef we can assume its a graph def format.

Alternatively we can be super lazy:
```
try:
  graph_def_import()
except GraphDefImportExceptionNotGraphDef:
  logger.info("Whoops not a graph def")
try:
  import_torch_from_pb()
except ...
...
```
XD

utensor_cgen/cli.py
dboyliao: I know.
I promise.

utensor_cgen/frontend/tensorflow.py
dboyliao: Honestly, I don't think it's a good solution.
As the community is concerned, the file extension convention is very well-known.

The intension of my decorator api is very explicit.
(Unfortunately I have to support python2 or I'll make `target_exts` a keyword-only argument)
So I think it would be easy for future developers/users to adapt to this approach.

How do you think? @Knight-X @neil-tan @mbartling

neil-tan: I think this is a case of a more robust error handling. The extension should be enough for us to select the frontend in the most cases. @mbartling's case should result in exceptions later in the parses. We can choose to either warn the users or fool-proof the error. We can do that in future improvement PRs perhaps?

dboyliao: Sure, we can do that.
I think we can do it with a cli option such as:
```
# by the file extension
utensor-cli convert whatever_graph --output-nodes=out1,out2 --frontend-parse=default
# for trying every possible parser and use the first match
utensor-cli convert whatever_graph --output-nodes=out1,out2 --frontend-parse=all
# or force to use specific parser
utensor-cli convert whatever_graph --output-nodes=out1,out2 --frontend-parse=tensorflow
```
just name a few.

I'll add it to issue tracker

PR #50, opened 2018-12-04
F/ram tensor

neil-tan:
Ram tensor can be generated by specifying op_type="Ram"
This is the minimal requirement for CMSIS-NN
We should look to extend this to static_add() and even inserting allocators operations in the graph


neil-tan:
Closing this pull request due to code contamination from the CMSIS-NN branch

PR #51, opened 2018-12-10
F/cmsis

neil-tan:
This is laying the foundation for future CMSIS-NN work/graph-rewriting functions to come.

Most of the functions are packaged in `ctensor_cgen/experimental`
The CMSIS-NN MatMul transformer is currently located in `ctensor_cgen/transformer/cmsis_nn.py`

More works need to be done for FC and Conv to be fully supported, but this is a good starting point.

The functions supplemented in the `ctensor_cgen/experimental` provides good insights to features that would be useful to uTensor code-generator.


dboyliao:
I can see it's a hack with no time.
I'm working on graph matcher at another branch.
The code should be cleaner with my implementation.

Since all the codes are in separate submodules, it should be easy to remove it afterward.
I'll merge it as is.

neil-tan:
See [PR53](https://github.com/uTensor/utensor_cgen/pull/53)

PR #57, opened 2019-02-24
Add id removal transformer

dboyliao:
Transformer which remove `Identity` op from Tensorflow


dboyliao:
close this PR.
duplicate #58

PR #63, opened 2019-03-25
Memory allocator pass

Knight-X:
1. create memory allocator in codegen with test cases
2. create a resource and allocation map for inspecting confliction
3. calculate the offset address in the runtime


tests/test_transformer/test_tensorallocator/test_allocator_optimizer.py
```diff
@@ -0,0 +1,127 @@
...
+    s = transformer._query_result(allocate_table, y, 6, 8, 1, 2)
+    assert s[0].name == 'y:0'
+    s = transformer._query_result(allocate_table, y, 9, 11, 1, 2)
+    assert s[0].name == 'y:0'
+    #address overlap, but time doesn't
+    s = transformer._query_result(allocate_table, y, 3, 6, 5, 6)
+    assert len(s) == 0
+    #time overlap, but address doesn't
+    s = transformer._query_result(allocate_table, y, 3, 4, 1, 2)
+    assert s[0].name == 'y:0'
```
neil-tan: Shouldn't this be `assert len(s) == 0` given it only overlaps in time?

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+        else:
+          new_end = max(end, new_end)
+      elif allocate_table[key]['offsetstart'] >= start and allocate_table[key]['offsetend'] >= start:
+        if allocate_table[key]['offsetend'] >= end:
+          new_end = max(new_end, allocate_table[key]['offsetend'])
+        else:
+          new_end = max(end, new_end)
+    return new_start, new_end
+
+  def _query_time_fromallocate_table(self, allocate_table, start, end):
```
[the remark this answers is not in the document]
Knight-X: yes, it returns the tensors which are overlapping.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+          in_o.alloc_address[1] = allocate_table[in_o.name]['offsetend']
+      out_t_infos = ugraph.ops_info[node_name].output_tensors
+      for out_o in out_t_infos:
+        if out_o.name in allocate_table:
+          out_o.alloc_address[0] = allocate_table[out_o.name]['offsetstart']
+          out_o.alloc_address[1] = allocate_table[out_o.name]['offsetend']
+    return ugraph
+
+
+  def _query_offset_fromallocate_table(self, allocate_table, start, end):
```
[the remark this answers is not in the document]
Knight-X: ok, that is fine.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+              break
+            else:
+              success = False
+        if success == False:
+          break
+      if success == True:
+        allocate_success = True
+    for node_name in ugraph.topo_order:
+      in_t_infos = ugraph.ops_info[node_name].input_tensors
+      for in_o in in_t_infos:
```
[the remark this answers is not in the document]
Knight-X: Every tensor should be recorded at allocate_table in topological order.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+    time_start = start
+    time_end = end
+    for key in allocate_table:
+      if allocate_table[key]['start'] >= start and allocate_table[key]['end'] <= end:
+        continue
+      elif allocate_table[key]['start'] <= start and allocate_table[key]['end'] >= start:
+        if allocate_table[key]['end'] >= end:
+          time_end = max(time_end, allocate_table[key]['end'])
+        else:
+          time_end = max(end, time_end)
```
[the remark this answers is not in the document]
Knight-X: no, the time_end is possibly longer than end, because the tensor lifetime would cross between start and end.

tests/test_transformer/test_tensorallocator/test_allocator_optimizer.py
Knight-X: ''' #address overlap, but time doesn't
    s = transformer._query_result(allocate_table, y, 3, 6, 5, 6)
    assert len(s) == 0
'''
it means that address overlap, but time dosen't collide.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+        #print("now the output tensor is %s" %outtensor.name)
+        if outtensor.name not in resource_table:
+          lifetime = dict()
+          lifetime['start'] = len_map[node_name]
+          lifetime['end'] = len_map[node_name]  
+          resource_table[outtensor.name] = lifetime
+
+    return resource_table
+  
+  def _create_topo_list(self, ugraph):
```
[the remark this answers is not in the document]
dboyliao: And in fact, this function can be written in python with one-liner as following
```python
my_map = dict([(name: index) for index, name in enumerate(ugraph.topo_order)])
```
Also, above one-liner is python2&3 compatible.
So I strongly recommend you to remove this function and give your variable a better name instead.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+          time_end = max(time_end, allocate_table[key]['end'])
+        else:
+          time_end = max(end, time_end)
+    return time_start, time_end
+
+  def _query_result(self, allocate_table, op, offset, length, timestart,
+  timeend):
+    ret = list()
+    for key in allocate_table:
+      '''print(key)
```
dboyliao: Please remove these debugging comments in your PR.
It's annoying to read and in fact `pdb` should be a better option in python.
We are not writing C/C++.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
 
 from .base import Transformer
 
-__all__ = ["DropoutTransformer", "BatchNormTransformer", "InlineTransformer"]
+__all__ = ["DropoutTransformer", "BatchNormTransformer", "InlineTransformer", "TensorLifeProbe"]
+class TensorLifeProbe(Transformer):
```
dboyliao: an empty newline between `__all__` and `class TensorLifeProbe` plz.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+    attribute['offsetstart'] = offset_start
+    attribute['offsetend'] = offset_end
+    allocate_table[tensor.name] = attribute
+    return allocate_table   
+
+  def _create_resource_table(self, ugraph):
+    resource_table = dict()
+    len_map = self._create_topo_list(ugraph)
+    for node_name in ugraph.topo_order:
+      #print("now is node %s, len is %s" %(node_name, len_map[node_name]))
```
dboyliao: These unnecessary `print`, remove them plz.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+      print(allocate_table[key]['end'])
+      '''
+      if (allocate_table[key]['offsetstart'] >= offset and allocate_table[key]['offsetstart'] <= offset + length) or (allocate_table[key]['offsetstart'] <= offset and allocate_table[key]['offsetend'] >= offset):
+        if (allocate_table[key]['start'] >= timestart and allocate_table[key]['start'] <= timeend) or (allocate_table[key]['start']
+        <= timestart and allocate_table[key]['end'] >= timeend):
+          ret.append(op)
+
+    return ret
+
+  def _check(self, ugraph, allocate_table, use_def_table, tensor, tensor_offset_start, tensor_offset_end):
```
dboyliao: Seems like you don't need a `ugraph` here?

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+    valid = False
+    timestart = use_def_table[tensor.name]['start']
+    timeend = use_def_table[tensor.name]['end']
+    offset, length = self._query_offset_fromallocate_table(allocate_table, tensor_offset_start, tensor_offset_end)
+    timestart, timeend = self._query_time_fromallocate_table(allocate_table, timestart, timeend)
+    a = self._query_result(allocate_table, tensor, offset, length, timestart, timeend)
+    if len(a) == 0:
+      return True
```
dboyliao:
```
if len(a) == 0:
    valid = True
return valid
```

I think single return principle should apply to **ANY** language, not only python.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -13,7 +13,175 @@
...
+        success = False
+        in_t_infos = ugraph.ops_info[node_name].input_tensors  # may be input tensor doest not need this
+        for in_o in in_t_infos:
+          for i in range(0, buffer_size, unit_size):
+            tensor_size = self._getSize(in_o.shape)
+            if self._check(ugraph, allocate_table, use_def_table, in_o, i, i + tensor_size):
+              self._create_allocate_table(allocate_table, use_def_table, in_o, i, i + tensor_size)
+              break
+            else:
+              success = False
```
dboyliao: Looks like after this loop, `success` is always `False`.
Is this what you want?

Also, thought not very well-known, there is `else` block for `for` loop in python.

PR #65, opened 2019-04-13
Exp matcher

dboyliao:


dboyliao:
@mbartling 
Basically you can skip `utensor_cgen/experimental`.
That's @neil-tan's matcher and I think you've read them before.

mbartling:
Excellent work! It's going to take me a bit to sort through this PR. It's pretty extensive :)

dboyliao:
@mbartling I know.
It takes me a while to pull these code off.
But I suggest not to go too deep into the code, just review the overall interface and pay attention to obvious mistake such as typo, styling or simple performance issue.

Feel free to ask me anything.

README.md
```diff
@@ -79,9 +79,33 @@ In graph theory terminology, they are `leaf` nodes of your graph.
 
 Run `utensor-cli convert --help` for detailed information.
 
-# Example
+# Use `utensor_cgen` as Library
+
+## Subgraph Isomorphic Matcher and Nodes Fusion
+
+With `uTensorGraphMatcher`, performing a subgraph isomorphic match, replace or manipulate the matched subgraph is just a few line of code!
```
mbartling: With `uTensorGraphMatcher`, performing common subgraph tasks such as isomorphic matching along with replacing or manipulating the matched subgraph(s) takes just a few line of code!

README.md
```diff
@@ -79,9 +79,33 @@ In graph theory terminology, they are `leaf` nodes of your graph.
...
+
+### Node Fusion
```
mbartling: For these few examples, you might want one or two sentences explaining what they user is looking for.

utensor_cgen/backend/code_generator.py
```diff
@@ -81,8 +81,6 @@ def formatter(name, kwargs):
     )
     quant_ugraph = self._transform_graph(ugraph,
                                          self.trans_methods)
-    from utensor_cgen.ir.misc.graph_viz import viz_graph
```
mbartling: Thank you for removing this :)

utensor_cgen/backend/operators.py
```diff
@@ -18,7 +21,7 @@ class OperatorFactory():
   def createOperatorSnippet(self, op_info, **kwargs):
     op_type = op_info.op_type
     if op_type not in self._operators:
-      err_msg = "unsupported op type in uTensor: {}".format(op_type)
+      err_msg = "unsupported op type in uTensor: {op.name}, {op.op_type}".format(op=op_info)
```
mbartling: This is just a thought, but we might consider creating a default "Unsupported Op" that takes on any unknown op. Rather than exiting out, we can have an option for `--ignore-unsupported` which will do it's best to continue and attempt to generate the code with a `ctx.push(new {{op.name}}(....));` bit.

utensor_cgen/backend/operators.py
```diff
@@ -549,10 +571,12 @@ def __init__(self, op_info, **kwargs):
     to_eval = parser.get('to_eval', False)
     self._snippet = QuantRangeForMultiplicationSnippet(inputs, outputs, output_type, ref_counts, to_eval)
 
+
 @OperatorFactory.register
+@OpEqualityDelegate.is_compatible_with("Const", _morphism.Inline2ConstMorphism)
```
mbartling: I like this format

utensor_cgen/transformer/conv_pool.py
```diff
@@ -85,6 +87,91 @@ def transform(self, ugraph):
...
+
+class ConvPoolTransformer(Transformer):
+  METHOD_NAME = 'conv_pool_v2'
+  KWARGS_NAMESCOPE = '_conv_pool_v2'
+
+  def __init__(self):
+    super(ConvPoolTransformer, self).__init__(prune_graph=False)
+
+  @property
+  def pattern_ugraph(self):
```
mbartling: Eventually it would be nice to figure out a better way of declaring the sub graphs. but the actual matching bit is so much nicer than it used to be

utensor_cgen/transformer/conv_pool.py
```diff
@@ -85,6 +87,91 @@ def transform(self, ugraph):
...
+      op_type='QuantizedFusedConv2DMaxpool',
+      backend=subj_conv_op.backend,
+      op_attr={
+        '_utensor_conv': subj_conv_op.op_attr,
+        '_utensor_pool': subj_pool_op.op_attr,
+      },
+      ugraph=repl_ugraph
+    )
+    topologic_order_graph(repl_ugraph)
+    input_map = {
```
mbartling: These lines can be procedurally done right?

utensor_cgen/transformer/linear_reoder.py
```diff
@@ -79,4 +80,57 @@ def transform(self, ugraph):
...
+class LinearReorderTransformerV2(Transformer):
+  METHOD_NAME = 'linear_reorder_v2'
+  KWARGS_NAMESCOPE = '_linear_reorder_v2'
+
+  def __init__(self):
+    self.prune_graph = False
+
+  @property
+  def pattern_ugraph(self):
+    graph = tf.Graph()
```
mbartling: Probably an insanely stupid question, but what does this subgraph have to do with LinearReorder?

README.md
dboyliao: Better.

README.md
dboyliao: Right. I'll elaborate it more.

utensor_cgen/backend/code_generator.py
dboyliao: No problem, bro.

utensor_cgen/backend/operators.py
dboyliao: I prefer `fail fast` in python.
Because it's the whole point of dev in python.
Fail fast, iterate faster.

utensor_cgen/backend/operators.py
dboyliao: Not sure if `morphism` is a good idea.
Not very useful now. Maybe it'll do some cool stuff later (?

utensor_cgen/transformer/linear_reoder.py
dboyliao: You mean the naming?
I just follow @neil-tan's naming.

utensor_cgen/transformer/conv_pool.py
dboyliao: I think it's non-trivial how to map the inputs/outputs of pattern graph and matched subgraph.
So I have to leave it to the users.

utensor_cgen/transformer/conv_pool.py
dboyliao: Sure.
But it'll like doing our own `Tensorflow` python api again.
Not of first priority for now.

utensor_cgen/backend/operators.py
mbartling: Assuming the morphisms can chain this could be super useful. Doesnt seem too dangerous

utensor_cgen/transformer/conv_pool.py
mbartling: Completely agree, let's just add this as a TODO feature in the issue tracker

utensor_cgen/transformer/linear_reoder.py
mbartling: @neil-tan The hell are you doing here? XD

dboyliao: I know this.
Just rearrange the layers to save memory.
`relu -> maxpool` is equivalent to `maxpool -> relu`
And @neil-tan call this linear_reorder.

utensor_cgen/transformer/conv_pool.py
dboyliao: Right. Will do.

README.md
neil-tan: We should make a note here that future transformers will be based on our own graph-constructors/builders.

utensor_cgen/backend/operators.py
neil-tan: Might be useful for stuff like architecture search.
However, we should also define the transform functions morphing from one graph to its equivalences. In addition to `op_type` we should also define how the tensor values need to be modified.

utensor_cgen/backend/operators.py
```diff
@@ -46,6 +55,9 @@ def snippet(self):
 
 
 @OperatorFactory.register
+@OpEqualityDelegate.is_associative(
+  permutations=((0, 1), (1, 0))
```
neil-tan: Does this support groupings? For example;
`q_add(a, a_min, a_max, b, b_min, b_max)` == `q_add( b, b_min, b_max, a, a_min, a_max)`

README.md
neil-tan: We should make a note here that future transforms will be based on our own graph constructor/builder.

utensor_cgen/backend/operators.py
neil-tan: Might be useful for stuff like architecture search.
However, we should also define the transform functions morphing from one graph to its equivalences. In addition to op_type we should also define how the tensor values need to be modified.
Seems doable with the current setup.

utensor_cgen/transformer/linear_reoder.py
neil-tan: @mbartling this is to reduce the memory needed to store the activation by taking advantage of down-sampling

utensor_cgen/backend/operators.py
dboyliao: I don't get it.
The later can be configured as `(3, 4, 5, 0, 1, 2)`.
Well, a tuple is kinda a formal math notation for permutation.

dboyliao: All you have to do is to list all possible permutations in this decorator.

utensor_cgen/transformer/conv_pool.py
dboyliao: https://github.com/uTensor/utensor_cgen/issues/80

utensor_cgen/backend/operators.py
dboyliao: Yes.
But I still have a gut feeling that the current setup may not be good enough.
Need a few iterations of refactoring. We'll see. : )

mbartling:
Note: This PR passes integration tests with the following PB file
[mnist_for_mc.tar.gz](https://github.com/uTensor/utensor_cgen/files/3382367/mnist_for_mc.tar.gz)

utensor_cgen/backend/operators.py
neil-tan: Ok. As long as this supports multiple permutations. It shouldn’t be a problem since the numbers of operants are usually small.

PR #66, opened 2019-04-27
added graphViz transformer

neil-tan:
usage: `utensor-cli convert simple_graph.pb  --output-nodes=u --transform-methods=dropout,quantize,inline,biasAdd,remove_id_op,refcnt,graphViz`

The can be place anywhere along the `--transform-methods` to instrument and inspect the graph.

### DIFF
--- a/utensor_cgen/ir/misc/graph_viz.py
+++ b/utensor_cgen/ir/misc/graph_viz.py
@@ -1,0 +1,31 @@
+from utensor_cgen.ir import OperationInfo, uTensorGraph
+from utensor_cgen.logger import logger
+from graphviz import Digraph
+
+def viz_graph(out_fname, view, ugraph):
+    dot = Digraph()
+    nodes = {}
+    i = 0
+    for node in ugraph.ops:
+        nodes[node.name] = chr(ord('a') + i)
+        dot.node(nodes[node.name], "%s: %s" % (node.name, node.op_type))
+        i += 1
+        for n in node.input_tensors:
+            if n.name in nodes:
+                continue
+            nodes[n.name] = chr(ord('a') + i)
+            dot.node(nodes[n.name], "%s: Tensor" % n.name)
+            i += 1
+        for n in node.output_tensors:
+            if n.name in nodes:
+                continue
+            nodes[n.name] = chr(ord('a') + i)
+            dot.node(nodes[n.name], "%s: Tensor" % n.name)
+            i += 1
+    for node in ugraph.ops:
+        for n in node.input_tensors:
+            dot.edge(nodes[n.name], nodes[node.name])
+        for n in node.output_tensors:
+            dot.edge(nodes[node.name], nodes[n.name])
+    dot.render(out_fname, view=view)
+    logger.info('graph visualization file generated: %s', out_fname)
--- a/utensor_cgen/transformer/graph_viz.py
+++ b/utensor_cgen/transformer/graph_viz.py
@@ -1,0 +1,25 @@
+# -*- coding:utf8 -*-
+r"""Graph Visualization Transformer
+
+Transformers that outputs a image file presenting the ugraph before the transformation
+"""
+import re
+from collections import defaultdict
+from copy import deepcopy
+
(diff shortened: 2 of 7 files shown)


utensor_cgen/transformer/graph_viz.py
```diff
@@ -0,0 +1,25 @@
...
+from .base import Transformer
+
+__all__ = ["GraphVizTransformer"]
+
+
+class GraphVizTransformer(Transformer):
+  METHOD_NAME = 'graphViz'
+  KWARGS_NAMESCOPE = '_utensor_graphViz'
+  
+  def transform(self, ugraph):
```
dboyliao: Two issues here:
1. As I said before, I think graph visualization should not be a part of `ugraph`. The main purpose of `ugraph` is to represent the structure and information needed to generate the NN code. We should refactor the viz part as another object or function which is responsible only for visualization. The other reason for doing it is that, if you take any library doing drawing/visualization as example, you'll find that there will be lots of parameters related to visualization. If we need to better visualize our `ugraph`, you can imagine how much parameters we can add to `ugraph` which are *irrelevant* to code generation if `ugraph` itself have to take care of visualization.
2. I suppose this transformer are designed to generate pictures of different stages of the `ugraph`, but it seems like you always overwrite the old pic with new one since they all share the same output file name. You need a way to write output to different files.

dboyliao:
@neil-tan 
I've patched your PR.
But it requires some changes to how we configure the pipeline.
The old way: `--transform-methods=dropout,quantize,inline,bias_add,remove_id_op,refcnt`
The new way: `--transform-methods=dropout|>quantize|>inline|>biasAdd|>remove_id_op|>refcnt`
In addition, we can specify keyword arguments as following:
`--transform-methods=dropout|>quantize|>inline|>graph_viz(out_fname="after_inline.gv")|>biasAdd|>remove_id_op|>graph_viz(out_fname="after_remove_id.gv", view=True)|>refcnt`

utensor_cgen/transformer/graph_viz.py
dboyliao: Issues solved with my patch.

neil-tan: @dboyliao    thanks for the improvement.
It looks like the `viz_graph()` is now part of the `Transformer` class
Any idea how we are going to visualize the graphs at different stages of a transform? (transformers other than `GraphVizTransformer`, as they won't inherit `viz_graph()`.

dboyliao: In order to keep things reasonable and maintainable, I think it's good to keep a transformer atomic.
As a result, if you want to visualize multiple stages within one transformer, it's a good sign that we should refactor the transformer into multiple transformers.
After that, you just insert `graph_viz` transformer after the transformer of the stage you want to visualize.

neil-tan: @dboyliao 
Transformers should be atomic. But, the graph viz is a powerful debugging tool. We often used it as the pdb for uGraph.

dboyliao: Right.
But for now I think a transformator can serve it pretty well.
If you want to make graph viz works like a debugger, it requires more refactoring.

utensor_cgen/ir/misc/graph_viz.py
```diff
@@ -0,0 +1,31 @@
+from utensor_cgen.ir import OperationInfo, uTensorGraph
```
dboyliao: Though after python 3.3, the `__init__.py` is no longer required.
(https://stackoverflow.com/questions/448271/what-is-init-py-for)
But according to the the official python reference, you are using a namespace package rather than regular package.
https://docs.python.org/3.7/reference/import.html#namespace-packages
https://www.python.org/dev/peps/pep-0420/

In our case, I recommend you to add `__init__.py` to make `misc` a regular python module.

dboyliao: BTW, python2 should not work with namespace package.

PR #67, opened 2019-05-03
F/fused conv gather

mbartling:


neil-tan:
Merging the approved changes to `develop`, continuing the graph rewriting work to facilitate node-fusion.

PR #69, opened 2019-05-06
F/quantizedfusedconv

mbartling:
Quantize Fused Convolution 2D Maxpool


mbartling:
time to self merge :D

PR #73, opened 2019-05-12
draft for data manager

Knight-X:
create a  data manager to handle different auxiliary information for generator

### DIFF
--- a/utensor_cgen/ir/instr.py
+++ b/utensor_cgen/ir/instr.py
@@ -6,6 +6,9 @@ class MemoryAllocation:
 
     def __setattr__(self,att,value):
         self._address_map[att] = value
+    
+    def __getattr__(self, att):
+        return self._address_map[att]
 
 class DataManager:
     InfoCenterMap = {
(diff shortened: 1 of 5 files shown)


utensor_cgen/ir/instr.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __setattr__(self,att,value):
+        self._address_map[att] = value
+    
+    def __getattr__(self, att):
+        return self._address_map[att]
+
+class DataManager:
+    InfoCenterMap = {
+        MemoryAllocation.store_name: MemoryAllocation
```
dboyliao: Should be `_store_name`?

utensor_cgen/utils.py
```diff
@@ -118,18 +118,25 @@ def parse_tensor_name(tname):
 
 class NamescopedKWArgsParser:
 
-  def __init__(self, name_space, kwargs):
+  def __init__(self, name_space, data_manager, op_info, kwargs):
```
dboyliao: For backward compatibility, it's better to make `data_manager` and `op_info` keyword arguments.
Or you'll fix all codes where use `NamescopedKWArgsParser`.

PR #74, opened 2019-05-27
rename transformer/pipline to transformer/pipeline

tung7970:
Rename transformer/pipline to transformer/pipeline


dboyliao:
Oh, thanks for fixing my typo.

PR #86, opened 2019-09-01
Mem allocator pr 

Knight-X:
1. memory allocation probe pass and generator

### DIFF
--- a/tests/test_transformer/test_tensorallocator/test_allocator_optimizer.py
+++ b/tests/test_transformer/test_tensorallocator/test_allocator_optimizer.py
@@ -135,25 +135,11 @@ def test_query_check(refgraph_tuple):
     assert valid == False
 
 def test_memory_allocation(refgraph_tuple):
-    #print("traditional way")
     (graph_def, output_nodes)= refgraph_tuple
     ugraph = GraphDefParser.parse(graph_def, output_nodes=output_nodes)
     transformer = TensorLifeProbe()
     ugraph = transformer.transform(ugraph)
-    for k, v in ugraph.data_manager.address.__dict__.items():
-        print(k)
-        print(v)
-    for node_name in ugraph.topo_order:
-      print("node {}".format(node_name))
-      in_t_infos = ugraph.ops_info[node_name].input_tensors
-      print("inputs")
-      for in_o in in_t_infos:
-        print(in_o.name)
-      out_t_infos = ugraph.ops_info[node_name].output_tensors
-      print("outputs")
-      for out_o in out_t_infos:
-        print(out_o.name)
-      print("next step")
+
     
     
 


utensor_cgen/ir/base.py
```diff
@@ -205,6 +206,7 @@ class uTensorGraph(IRBase, _NoShallowCopyMixin):
   _backend = attr.ib(default='', type=str)
   ops_info = attr.ib(factory=dict)
   topo_order = attr.ib(factory=list, init=False)
+  data_manager = None
```
dboyliao: It should be `data_manager = attr.ib(default=None)`.
Or `data_manger` will be a shared class-level attribute.

Please take a look over these examples for `attrs`:
http://www.attrs.org/en/stable/examples.html

utensor_cgen/ir/base.py
```diff
@@ -267,4 +273,5 @@ def __deepcopy__(self, memo):
     new_graph.ops_info = new_ops_info
     new_graph.topo_order = new_topo_order
     new_graph._backend = self._backend
+    new_graph.data_manager = self.data_manager
```
dboyliao: You should make a copy of `data_manager` not just pass reference to new graph.
This violate the semantic of `deepcopy`.

dboyliao: https://docs.python.org/3.6/library/copy.html

utensor_cgen/ir/instr.py
```diff
@@ -0,0 +1,53 @@
+class MemoryAllocation:
```
dboyliao: You should make the base class as `object` or `MemoryAllocation` is a old style class which has different typing system as new style class in python2.

utensor_cgen/ir/instr.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __setattr__(self, attr, value):
+        if attr == '_address_map':
+            return super().__setattr__(attr, value)
+        self._address_map[attr] = value
+    
+    def __getattr__(self, att):
+        return self._address_map[att]
+
+class DataManager:
```
dboyliao: Missing the base class `object`.

utensor_cgen/ir/instr.py
```diff
@@ -0,0 +1,53 @@
+class MemoryAllocation:
+    _store_name = "address"
+    
+    def __init__(self):
+        self._address_map = {}
+
+    def __setattr__(self, attr, value):
+        if attr == '_address_map':
+            return super().__setattr__(attr, value)
```
dboyliao: It should be `super(MemoryAllocation, self)`

utensor_cgen/ir/instr.py
dboyliao: https://docs.python.org/2.5/ref/node33.html

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -9,11 +9,187 @@
 from copy import deepcopy
 
 from utensor_cgen.ir import OperationInfo, uTensorGraph
-from utensor_cgen.utils import parse_tensor_name
-
+from utensor_cgen.utils import parse_tensor_name 
 from .base import Transformer
+import pdb
```
dboyliao: remove `pdb` import.

utensor_cgen/ir/instr.py
```diff
@@ -0,0 +1,53 @@
...
+        self.StorageCenter = {}
+        for data_name in datas:
+            data_cls = self.InfoCenterMap.get(data_name, None)
+            if data_cls is None:
+                raise ValueError("Unknown transformation method: {}".format(data_name))
+            datastorage = data_cls()
+            self.StorageCenter.update({data_name: datastorage})
+
+    @classmethod
+    def register_datamap(cls, data_cls, overwrite=False):
```
dboyliao: why do you need this `overwrite` arguments?
Looks like it's of no use in the function body.

utensor_cgen/ir/instr.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __getattr__(self,attr):
+        if attr == 'StorageCenter':
+            raise AttributeError('StorageCenter')
+        cls_instance = self.StorageCenter[attr]
+        return cls_instance
+
+    def __setattr__(self,attr,value):
+        if attr == 'StorageCenter':
+            return super().__setattr__(attr, value)
```
dboyliao: `super(DataManager, self)`

utensor_cgen/ir/instr.py
dboyliao: `super` works differently in python2 and python3.

utensor_cgen/transformer/ns_transformer.py
```diff
@@ -9,11 +9,187 @@
...
+  METHOD_NAME = 'tensorlife'
+  KWARGS_NAMESCOPE = '_utensor_utlife'
+  DATA_NAME = 'address'
+
+
+  def transform(self, ugraph):
+    ugraph.create_data({self.DATA_NAME: " "})
+    use_def_table = self._create_resource_table(ugraph)
+    unit_size = 4
+    buffer_size = 100000 #1k bytes
```
dboyliao: Why it's not 1024 (`1k bytes`)?

utensor_cgen/ir/instr.py
```diff
@@ -0,0 +1,53 @@
...
+        cls_instance = self.StorageCenter[attr]
+        k, v= value
+        cls_instance.__setattr__(k, v)
+
+    def group(self, tensor):
+        ret = {}
+        for key, cls_object in self.StorageCenter.items():
+            ans = cls_object.__getattr__(tensor)
```
dboyliao: https://docs.python.org/3/reference/datamodel.html#object.__getattr__

This violates the definition of `__getattr__` in python. 
`__getattr__` should take a name which is a `str` and return the attribute of the object.

utensor_cgen/ir/base.py
Knight-X: if I use data_manager = attr.ib(default=None), should I have to deepcopy it in the new graph?

dboyliao: Depends on how you'd like to use the data_manager.
If you don't want to share states among old and new graphs, you should make a deepcopy.

PR #88, opened 2019-09-08
Mem allocator

Knight-X:


dboyliao:
mem alloc optimization pass is optional now.

PR #92, opened 2019-11-26
bugfixes:

neil-tan:
- setup.py: fixed onnx-tf version and syntax touch up for graphviz
- syntax fix for QntMatMulOp template file

note: current onnx-tf 1.5x throws error


dboyliao:
looks good to me

PR #93, opened 2019-11-29
Tflite backend

neil-tan:
This is semi-working for:

input0 = [1,1,1,1].tranpose    #set at the run time
input1 = [2,4,6,8].tranpose
weight = [10,20,30,40]
bias = [7]

add_output = ADD(input0, input1)
out = FC(add_output, weight, bias)

out == 707

Unexplained behaviors: when the first tensor in the subgraph, `input0`, is not set at the runtime, the tensor is then initialized to undefined values instead of values contained in FlatBuffers. See [here](https://github.com/uTensor/utensor_cgen/blob/c034f7de86d3ebf33bb39ccc59a550b03eb9e77e/tests/tflm/tflite_export/test_write.py#L98).

@dboyliao   Can use some input on the python coding style/design patterns. Keep in mind, I barely got this to work.


neil-tan:
@dboyliao The first check passed, for you to take it for a spin.

utensor_cgen/third_party/flatbuffers/builder.py
```diff
@@ -0,0 +1,764 @@
+# Copyright 2014 Google Inc. All rights reserved.
```
dboyliao: WTF @neil-tan, do you copy the entire flatbuffers python library in this PR!?
Why do you do this?

utensor_cgen/transformer/tflite_exporter.py
```diff
@@ -0,0 +1,412 @@
...
+  schema_version = 3
+  file_ident = "TFL3"
+
+  def __init__(self, input_tensors, output_tensors):
+    """
+    input_tensors: a list of input tensor names
+    output_tensors: alist of output tensor names
+    """
+
+    self.prune_graph = False #what is this?
```
dboyliao: I did some trick on the `Transformer`.
And setting this flag to `False` simply to skip a pruning phase.
Or all transformed will go through a pruning phase to check if there is any op that is no longer used in the graph and it will remove such ops.

utensor_cgen/transformer/tflite_exporter.py
```diff
@@ -0,0 +1,412 @@
...
+      preprend_method(it)
+    
+    return self.fbuilder.EndVector(num_items)
+
+  def __tflm_graph_legalize_(self, ugraph):
+    for _, op in ugraph.ops_info.items():
+      assert op.op_type.startswith("TFLM_") or op.op_type in self.static_op_types
+      #TODO: strip the leading TFLM_ only
+      op.op_type = op.op_type.replace("TFLM_", '')
+# How to construct the quantization parameter for intermediate tensors?
```
dboyliao: I don't know what you mean by quantization parameters for intermediate tensors

utensor_cgen/third_party/flatbuffers/builder.py
neil-tan: The current version of Pypl flatbuffers package is missing the ability to add an identifier string to the model file. So, a copy of the latest flatbuffer (github) python api is included. The generated tflm Flatbuffer api, however, reference the standard import.

It would be great if we can make the latest flatbuffer importable, but I had to settle for Tods workaround now.

utensor_cgen/transformer/tflite_exporter.py
neil-tan: This is a misplaced comment meant for the routine that construct the quantization parameters for intermediate tensors. It’s likely to have something to do with offline quantizations.

utensor_cgen/third_party/flatbuffers/builder.py
dboyliao: I can't use github repo as dependency since the repo is not setuptools compatible.
Fuck that.

Though I can write an install hook which will clone and install latest version from the repo, that will make the install process slow.
My suggestion is copy the core library only and remove all auto generated files since that could be generated during installation.

Fuck, I'm so pissed.

neil-tan: @dboyliao good proposal generating the TFLM binding as a part of the setup process btw, as discussed offline yesterday.

dboyliao: Yay, I'll help you with that after merging #94.
BTW, I add legalization pass in the TF frontend.
It's part of the frontend but this is temporary.
I will refactor the legalization and lowering as separate phases for codegen in the future.

neil-tan: Splendid

dboyliao:
BTW, I think it's not proper to make it a tflm transformer.
It's better to implement it as a backend.

neil-tan:
Nope, it isn't. This was prototyped as a transformer as the backend was being refactored. Changes need to be migrated.

dboyliao:
Yes, I know.
I just want to leave a note so we'll not forget to fix it before merge this PR.
Should not hard converting it into a backend.

PR #94, opened 2019-12-13
backend refactoring

dboyliao:
Backend Refactoring

In this PR, I define the `Backend` interface and the entire code generation is simplified as:
1. select a proper frontend parser and parse the input model file as IR graph
2. pass the IR graph to target backend

As a result, if we want to support different output format, we only need to implement different backend.
Here is the cli code if you want to take a look an example:
https://github.com/uTensor/utensor_cgen/blob/backend-refactor/utensor_cgen/cli.py#L68-L75


utensor_cgen/backend/utensor/_graph_lower.py
```diff
@@ -0,0 +1,23 @@
...
+from utensor_cgen.utils import class_property
+
+
+class uTensorGraphLower(BackendPart):
+
+  TARGET = 'utensor'
+  PART = 'graph_lower'
+
+  def apply(self, ugraph):
+    handler = getattr(self, 'handle_{}'.format(ugraph.lib_name))
```
neil-tan: All backend should be mandated to implement a common uTensor IR handler. Otherwise, we cannot share optimization passes between different front/backend configurations. Of course, additional front/backend handlers can be added, but these are secondary and optional to the uTensor IR.

I'll be opening an issue to add a prefix `uTensor_` to all uTensor operator names (ones that actually gets to JinJa2). Then, start defining the uTensor IR. We have the option to adapt ONNX, TF Lite or our own.

dboyliao: I don't think add prefix is a good solution.
I'm thinking a more robust way such as add a legalization phase to the cli.

Graph legalization should be responsible to convert all platform dependent ops to generic IR ops.
And we implement our optimizers, matchers, ...etc against the generic IR graph.
Then in graph lowering phase, we convert generic IR graph to target dependent graph.

However, graph legalization is not included in this PR and hence the graph lowering here is like experimental and requires further refactoring.
I add lowering in this PR is just a temporary workaround on the issue (pytorch, tflite-micro, ...etc) we talk about few days ago.

PR #102, opened 2020-02-26
Onnx Frontend

dboyliao:
1. Onnx frontend
    - core logic to review: https://github.com/uTensor/utensor_cgen/blob/onnx/utensor_cgen/frontend/onnx.py#L38-L52
2. Offline Tensor Allocation via Or-Tools
    -  the code: https://github.com/uTensor/utensor_cgen/blob/onnx/utensor_cgen/backend/graph_lower/generic_graph_lower.py#L63-L252

Those are the key bits of this PR.


utensor_cgen/backend/graph_lower/alloc_plan.py
```diff
@@ -0,0 +1,112 @@
...
+  def __contains__(self, slot):
+    assert isinstance(slot, int), 'incorrect slot type: %s' % type(slot)
+    is_in = self.time_slot_start <= slot
+    if self.time_slot_end is not None:
+      is_in = is_in and slot <= self.time_slot_end
+    return is_in
+
+
+@attr.s
+class SpaceAllocation(object):
```
mbartling: This will need to be able to account for memory access alignment. Probably need to include this info in the attributes

utensor_cgen/backend/transformer.py
```diff
@@ -0,0 +1,74 @@
...
+    self.transformer = TransformerPipeline(
+      methods=self.config['transform_methods']
+    )
+    self.trans_methods = self.config['transform_methods']
+    self.save_graph = self.config['save_graph']
+
+  def transform(self, ugraph):
+    logger.info("Transforming graph: %s", ugraph.name)
+    logger.info("Transform pipeline: %s", ' -> '.join(self.trans_methods))
+    self._check_non_quantized(ugraph)
```
mbartling: scandalous

utensor_cgen/backend/transformer.py
```diff
@@ -0,0 +1,74 @@
...
+      with open(pkl_fname, 'wb') as fid:
+        pickle.dump(new_ugraph, fid)
+      logger.info('{} saved'.format(pkl_fname))
+    return new_ugraph
+  
+  @classmethod
+  def _check_non_quantized(cls, ugraph):
+    is_quantized = False
+    quant_ops = set([
+      "Dequantize", "QuantizedMaxPool",
```
mbartling: These will be the uTensor equivalent names correct?

utensor_cgen/backend/utensor/_backend_impl.py
```diff
@@ -1,39 +1,45 @@
...
-uTensorRearchGraphLower = LazyAttrib(_graph_lower, 'uTensorRearchGraphLower')
+uTensorLegacyGraphLower = LazyAttrib(_op_lower, 'uTensorLegacyGraphLower')
+uTensorRearchGraphLower = LazyAttrib(_op_lower, 'uTensorRearchGraphLower')
+TensorAllocationPlanner = LazyAttrib(generic_graph_lower, 'TensorAllocationPlanner')
+PipelineTransformer = LazyAttrib(transformer, 'PipelineTransformer')
 
-del code_generator, _graph_lower
+del code_generator, _op_lower, generic_graph_lower
 
 class uTensorBackend(Backend):
```
mbartling: This class is SUPER important as an example backend, but it is missing some key docs like how it links the middleware bits to the target backend.

utensor_cgen/backend/utensor/_graph_lower/_op_lower.py
```diff
@@ -25,11 +31,10 @@ def default_config(cls):
     return {}
 
 
-class uTensorRearchGraphLower(BackendPart):
-  TARGET = 'utensor'
+class uTensorRearchGraphLower(uTensorGraphLowerBase):
   PART = 'rearch_graph_lower'
 
-  class OptypRenameManager(object):
+  class OptypeRenameManager(object):
```
mbartling: Something that might be useful is a specification for the ops on the backend that describe names, inputs, outputs, options (passed to the constructor) and potentially generic knowledge of sizes/validation functions. We kind of already have this so it might be easy to link in here vs the select op stage later.

README.rst
```diff
@@ -226,12 +226,32 @@ Use Case: Dropout Layer Removal
...
+\ |mlp-alloc-graph|
+
+Once enabled the optimization transformer, ``tensor_alloc``,
+which is an offline tensor memory allocation planner,
+``utensor-cli`` will generate ``uTensor`` runtime codes that use following optimized allocation plan:
+
+\ |mlp-alloc|
+
+- On the y-axis, these are tensor names ordered by topological sorting
+- On the x-axis, these are the memory span occupied by each tensor
```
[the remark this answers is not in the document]
dboyliao: How about simply `memory`?
I'm not sure what is the right terminology.

example_config.toml
```diff
@@ -1,15 +1,50 @@
 # https://github.com/toml-lang/toml
 # <target_name>.<component>.<part>
+# we use string 'None' to represent python None value
+# you should convert the string to None if you try to write extension for utensor_cgen
 [utensor.backend]
 legacy-api = true
 
+[utensor.backend.tensor_alloc_planner]
+max_pool_size = 1048576
```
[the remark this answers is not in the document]
dboyliao: I did add a comment in the code.
It was `1024*1024` which is 1 GB.
Setting it a large number just gives the planner enough free space for a feasible solution.
With a least one feasible solution, we are able to tell the user the minimum required memory for running a graph.

utensor_cgen/backend/base.py
```diff
@@ -1,7 +1,7 @@
-from utensor_cgen.utils import (
-  MUST_OVERWRITEN, class_property,
-  parse_toml, Configuration,
-)
+from abc import abstractmethod
+
+from utensor_cgen.utils import (MUST_OVERWRITEN, Configuration, class_property,
```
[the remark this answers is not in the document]
dboyliao: Oh, right.
A typo.
Auto completion makes me ignore this.
I'll make it `MUST_OVERWRITE`.

utensor_cgen/backend/transformer.py
dboyliao: I think it's better to make sure users pass a non quantized graph to the transformer.
If quantization is required, it's better to be done by the transformer backend part.
Not sure if you have other issue in mind.

utensor_cgen/backend/transformer.py
dboyliao: Well, yes.
I'm thinking we will need to formally define a set of op names for quantization (generic names).
We don't have such generic ops yet but it's not hard to fix it once we do.

utensor_cgen/backend/utensor/_backend_impl.py
dboyliao: Right, I'm about to update the architecture of the cli in the docs.
By the time the backend is written, I'm trying to come up with a proper way to do it so I didn't write any docs.

utensor_cgen/backend/graph_lower/alloc_plan.py
dboyliao: I add one more config field, `data_alignment`, for the `TensorAllocationPlanner`.
Now the planner is aware of data alignment and it's configurable to account for different platform.
It's default to 2-bytes alignment.
I'll add this attribute to `SpaceAllocation`.

dboyliao: done

utensor_cgen/backend/utensor/_graph_lower/_op_lower.py
dboyliao: According to my understanding, you are saying that we shift the checks from the code generator to the lowering phase right?
It's doable but I think it need some refactoring work to do.
I'll put it in the issue tracker. How do you think?

README.rst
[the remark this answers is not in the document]
mbartling: I mean it's kind of a generic term here. Might be good to specify range of memory addresses/offsets that span tensor.

example_config.toml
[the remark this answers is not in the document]
mbartling: kk

utensor_cgen/backend/graph_lower/alloc_plan.py
mbartling: marvelous :)

README.rst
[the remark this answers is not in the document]
dboyliao: I see. 
It do make things clearer.
Will do.
Maybe not this PR.
I found some of visualization parameters need some twists which I think is not very important for now.

utensor_cgen/backend/transformer.py
mbartling: We wont be able to guarantee this with quantized aware training. So it's probably best if we try to bake these checks in now.

utensor_cgen/backend/transformer.py
mbartling: We can always go off of the ops in the runtime, for naming requirements, in/outs, params, and additional attributes. Then for uTensor targets the conversion is a nop, for any other targets we just do some finagling

utensor_cgen/backend/utensor/_graph_lower/_op_lower.py
mbartling: We should probably discuss this offline first, there is a lot in this tiny post :)

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
```diff
@@ -26,26 +26,19 @@ def __init__(self, config):
...
-      with open('transformed_{}.pkl'.format(ugraph.name), 'wb') as fid:
-        pickle.dump(new_ugraph, fid)
     # 1. find all ops required
     ops = set()
     placeholders = set()
     tensor_var_map = {} # tensor name -> var name
-    for op_info in new_ugraph.ops_info.values():
+    for op_info in ugraph.ops_info.values():
       for tensor in op_info.output_tensors:
         tensor_var_name = re.sub(r'[:/]', '', tensor.name)
```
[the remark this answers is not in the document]
dboyliao: Yes, this should be in the frontend.
I'll fix it.

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
```diff
@@ -26,26 +26,19 @@ def __init__(self, config):
     self.src_fname = final_config['src_fname']
     self.header_fname = final_config['header_fname']
     self.params_dir = final_config['params_dir'].rstrip('/')
-    self.trans_methods = final_config['transform_methods']
     self.meta_data_pool_size = final_config['meta_data_pool_size']
     self.ram_data_pool_size = final_config['ram_data_pool_size']
     self.model_dir = final_config['model_dir'].rstrip('/')
-    self.save_graph = final_config['save_graph']
 
   def apply(self, ugraph):
```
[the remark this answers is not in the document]
dboyliao: You mean the comments?

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
```diff
@@ -63,7 +56,7 @@ def apply(self, ugraph):
       ops_map[op] = op_var_name
       declare_snippets.append(op.get_declare_snippet(op_var_name))
     weight_snippets = []
-    for op_info in filter(lambda op_info: op_info.op_type == 'Inline', new_ugraph.ops_info.values()):
+    for op_info in filter(lambda op_info: op_info.op_type == 'Inline', ugraph.ops_info.values()):
       tensor = op_info.output_tensors[0]
       buffer_name = 'data_{}'.format(tensor.name.replace(':', '_').replace('/', '_'))
```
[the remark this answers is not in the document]
dboyliao: No, it's not legalization.
It the variable name used in the weight header for referencing the buffer.

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
```diff
@@ -94,13 +87,13 @@ def apply(self, ugraph):
       )
     template_vars = {}
     template_vars['model_name'] = ugraph.name
-    template_vars['meta_data_pool_size'] = self._compute_meta_data_size(new_ugraph)
-    template_vars['ram_data_pool_size'] = self._compute_ram_data_size(new_ugraph)
+    template_vars['meta_data_pool_size'] = self._compute_meta_data_size(ugraph)
```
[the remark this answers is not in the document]
dboyliao: Ok, sounds reasonable.
I'll modify the method signature.
But to be honest, this method is experimental.
The rearch code generator will go for a major refactoring in the future for sure.

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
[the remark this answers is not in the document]
mbartling: I mean it might make things clearer with one subroutine for each chunklet For example:

`(ops, placeholders, tensor_var_map) = self._find_required_ops(...)`

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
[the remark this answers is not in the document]
mbartling: thanks!

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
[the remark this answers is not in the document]
mbartling: i agree, but even as a hack fix for now, keeping it in the codebase record will reduce the number of errors we make down the road.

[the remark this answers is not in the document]
dboyliao: Sure thing.
I'm just saying the function body might looks dumb for now. XD

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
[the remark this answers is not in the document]
dboyliao: Oh, no wait.
It's not legalization either.
The `tensor_var_name` is the variable name used in the generated source file.
Just as the `buffer_name` you mention in the other comment below.

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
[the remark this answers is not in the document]
dboyliao: Right. I see.
Sure thing.
Let's me think about what are the proper names for these subroutines.

utensor_cgen/backend/base.py
```diff
@@ -16,15 +16,33 @@ def __new__(cls, config, *args, **kwargs):
         validator(config, *args, **kwargs)
     if isinstance(config, dict):
       config = Configuration(cls.default_config, config)
+    elif config is None:
+      config = Configuration(cls.default_config, {})
     self._config = config
     return self
 
+  @abstractmethod
   def apply(self, ugraph):
```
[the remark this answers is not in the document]
dboyliao: You mean sth like this: `register_visit_op(op_type, callback)`.
So user can register a callback to the backend whenever some specific op_type is visited in the body of `apply`, the callback will be invoked?

And how does tensor visitor works? register a callback on tensor of specific data type?
Since tensor itself is kinda generic. It's functionality is mostly given by the op which owns it.

utensor_cgen/backend/transformer.py
dboyliao: How about we make exceptions to quantized aware trained graph?
I think there should be some unique ops used exclusive to quantization aware trained graph, right?
Or we can add special attribute value to quantization aware graph.

utensor_cgen/backend/transformer.py
dboyliao: Can you elaborate more about go off the ops in the runtime for naming requirements, in/outs, params and attributes? I don't understand what you mean.
An example will help.

utensor_cgen/ir/misc/graph_viz.py
```diff
@@ -65,3 +70,82 @@ def viz_graph(ugraph, out_fname=None, view=False, cleanup=True, colored_nodes=No
...
+        title = 'Optimal Tensor Allocation: {} bytes in total'.format(alloc_plan.total_size)
+      plt.title(
+        title,
+        fontdict={'fontsize': fontsize}
+      )
+      if figsize is None:
+        figsize = (num_tensors_per_split, num_tensors_per_split / 2)
+      fig.set_size_inches(*figsize)
+      fig.tight_layout()
+      figs.append(fig)
```
mbartling: Why not use subfigs?

utensor_cgen/legalizer/onnx.py
```diff
@@ -0,0 +1,57 @@
...
+from utensor_cgen.utils import topologic_order_graph
+
+from .base import LegalizerBase
+
+
+class OnnxLegalizer(LegalizerBase):
+  TARGET = 'onnx'
+
+  def legalize(self, ugraph):
+    self._visit(ugraph)
```
mbartling: A bit weird use of name visit, normally `visit` calls are on individual nodes, and not the whole graph. Consider using a `visit_all` method or similar

utensor_cgen/legalizer/onnx.py
```diff
@@ -0,0 +1,57 @@
...
+    for op_info in list(ugraph.ops_info.values()):
+      visitor = getattr(
+        self,
+        '_visit_{}'.format(op_info.op_type.lower()),
+        lambda op_info: op_info
+      )
+      visitor(op_info)
+    topologic_order_graph(ugraph)
+
+  def _visit_gemm(self, op_info):
```
mbartling: Where do these functions get called from, I noticed this pattern used elsewhere but it's a bit unclear how they're used

utensor_cgen/cli.py
```diff
@@ -35,12 +35,12 @@ def _load_plugin(path):
   "-p",
   "--plugin",
   multiple=True,
-  help="path of the python module which will be loaded as plugin",
+  help="path of the python module which will be loaded as plugin (multiple)",
```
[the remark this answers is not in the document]
dboyliao: you can pass values to this flag multiple time.
ex: `utensor-cli -p path1 -p path2 -p path3 ....`

utensor_cgen/frontend/onnx.py
```diff
@@ -1,19 +1,242 @@
...
+        name=self._format_tensor_name('', node_name, 0),
+        op_name=node_name,
+        dtype=tensor_value.value.dtype,
+        shape=list(tensor_value.value.np_array.shape),
+        ugraph=ugraph
+      )
+      OperationInfo(
+        name=node_name,
+        input_tensors=[],
+        output_tensors=[tensor_names_map[tensor_name]],
```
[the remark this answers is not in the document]
dboyliao: These ops is in the parameters list in onnx graph.
So these will be `Const` ops which has only one output for sure.

utensor_cgen/frontend/onnx.py
```diff
@@ -1,19 +1,242 @@
...
+      if tensor_name in tensor_names_map:
+        # tensor in initializers MAY appears in input, ignore
+        continue
+      cnt = op_types_cnt['Placeholder']
+      node_name = self._format_node_name(tensor_name, 'Placeholder', cnt)
+      op_types_cnt['Placeholder'] += 1
+      assert value.type.HasField('tensor_type'), 'invalid graph input value'
+      tensor_type = value.type.tensor_type
+      dtype_str = TensorProto.DataType.Name(tensor_type.elem_type).lower()
+      if dtype_str == 'float':
```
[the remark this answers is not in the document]
dboyliao: https://github.com/onnx/onnx/blob/master/onnx/onnx-ml.proto3#L452-L478

Pretty sure `float` is not `double` in ONNX.

utensor_cgen/frontend/onnx.py
```diff
@@ -1,19 +1,242 @@
...
+      output_nodes=[],
+      lib_name='onnx',
+      ops_info={},
+    )
+    self._build_graph(onnx_graph, ugraph)
+    ugraph = Legalizer.legalize(ugraph)
+    tf.reset_default_graph()
     return ugraph
+  
+  def _build_graph(self, onnx_graph, ugraph):
```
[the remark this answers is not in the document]
dboyliao: Ok, will do.

utensor_cgen/frontend/onnx.py
```diff
@@ -1,19 +1,242 @@
...
+
+  @classmethod
+  def post_process(cls, ugraph):
+    for op_name in ugraph.topo_order:
+      op_info = ugraph.ops_info[op_name]
+      handler = getattr(cls, '_handle_{}'.format(op_info.op_type.lower()), lambda op_info: op_info)
+      handler(op_info)
+
+  @staticmethod
+  def _handle_gemm(op_info):
```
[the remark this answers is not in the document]
dboyliao: In ONNX, GEMM = Matmul + BiasAdd.
I need to do some simple graph rewrite and renaming.
Also, the shape information is not recored in the ONNX graph.
And I find it's a good case to implement visit pattern so I implement the post processing like this.
Basically post processing part will do:
1. graph rewrite, or
2. renaming, or
3. finding shape and data types for inputs/outputs

utensor_cgen/ir/misc/graph_viz.py
dboyliao: You are right.
I just didn't think about it. XD

utensor_cgen/legalizer/onnx.py
dboyliao: Yes, that's right.
Will rename it as `visit_all`

utensor_cgen/legalizer/onnx.py
dboyliao: https://github.com/uTensor/utensor_cgen/pull/102/files/7d9d6eaeb9e380d233580612cb173e2ea7085e77#diff-78a1f0eaee7e58baa9a6b8b7c7954f74R16-R21

I use dynamic attribute lookup in python.
The `getattr` function.

utensor_cgen/ir/misc/graph_viz.py
```diff
@@ -1,7 +1,12 @@
-from graphviz import Digraph
+from random import random, seed
 
+import matplotlib.pyplot as plt
+from graphviz import Digraph
```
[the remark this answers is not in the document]
dboyliao: Right, it might take me some time to make the import optional.
Not a big issue I think.
How about put it in issue tracker?

utensor_cgen/ir/misc/graph_viz.py
mbartling: lolol I have a massive amount of experience with matlab so this is in my blood XD

dboyliao: Ok, I see.
BTW, I really hate matlab plotting api.
I swear to gods I'll replace `matplotlib` with `ggplot2` once I get a chance.

mbartling: both work, I have more experience with matlab style plots but I have no issues with switching

utensor_cgen/frontend/onnx.py
```diff
@@ -1,19 +1,242 @@
...
+        input_tensors=[],
+        output_tensors=[tensor_names_map[tensor_name]],
+        op_type='Placeholder',
+        ugraph=ugraph,
+        lib_name='onnx',
+        op_attr={}
+      )
+    for node in onnx_graph.node:
+      cnt = op_types_cnt[node.op_type]
+      node_name = self._format_node_name(node.name, node.op_type, cnt)
```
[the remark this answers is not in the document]
dboyliao: Not quite.
For onnx, node name might be missing (empty string) whereas tensorflow make node name as required.
So `_format_node_name` works a bit differently in these two parsers.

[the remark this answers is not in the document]
dboyliao: I think it's better to add a check in the base parser class.
And use some trick, like what I did for pruning graph in transformer, to make sure the check is ran after every call to `parse`.

utensor_cgen/backend/base.py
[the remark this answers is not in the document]
dboyliao: https://github.com/uTensor/utensor_cgen/issues/106

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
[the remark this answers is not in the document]
dboyliao: Done.

utensor_cgen/frontend/onnx.py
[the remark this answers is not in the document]
mbartling: 👍

utensor_cgen/frontend/onnx.py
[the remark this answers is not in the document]
mbartling: ok

utensor_cgen/legalizer/onnx.py
mbartling: Might consider adding this as a mini comment above, would be super helpful

utensor_cgen/backend/transformer.py
mbartling: Yeah right now my concern is for the various quantization frameworks. TF and PyTorch for example either make it part of training and then rewrite the graph, or do offline quantization and pass in representative inputs to compute the output ranges of the various operator nodes. Supporting all these runtimes might be a pain.

dboyliao: I have no good idea for now.
It's better to work with an example.
Current workflow expecting a non-quantized graph since many transformers are designed for non-quantized graph.
I think it's better to leave it as is for next release and see what is the better way to do it then.
Just for now.

utensor_cgen/legalizer/onnx.py
dboyliao: Ok

dboyliao: Done

utensor_cgen/backend/transformer.py
mbartling: @neil-tan thoughts? I can see reasons for both ways for now

utensor_cgen/legalizer/onnx.py
mbartling: <3

dboyliao:
closing this PR since it's merged into #108

PR #103, opened 2020-03-01
Ortools

dboyliao:
demo_mem_alloc

Experimental tensor allocation planner with [ortools](https://developers.google.com/optimization).

The picture above:
- the y axis are the tensor names sorted according to topological order of a mlp, from top to bottom.
- the x axis are the memory span of each tensors in the memory pool.


dboyliao:
@neil-tan @Knight-X @mbartling 
If you are interested in this PR, here is the key bits:
https://github.com/uTensor/utensor_cgen/pull/103/files#diff-ad6830d2f958b649ebbb340c55ab59e8R28-R252

PR #108, opened 2020-05-01
Re arch support

dboyliao:
re-arch code generator


utensor_cgen/frontend/tflite.py
```diff
@@ -0,0 +1,398 @@
...
+
+      input_tensor_names = [
+        self.tensor_names_map[input_index] for input_index in op.InputsAsNumpy()
+      ]
+      output_tensor_names = [
+        self.tensor_names_map[output_index]
+        for output_index in op.OutputsAsNumpy()
+      ]
+
+      op_attr = _OP_DATA_FUNC_MAP[op_type](op)
```
mbartling, line 361: Please throw error if op_type not in this table, else do best effort default op maybe? We might be able to `import_lib` and do a best effort reflection scan to fill out the `option_dict`. I feel like the frontend should support all ops as best effort, and the actual `is_op_supported()` call should be triggered from the backend

utensor_cgen/backend/transformer.py
```diff
@@ -0,0 +1,75 @@
...
+    if self.save_graph:
+      logger.info('Saving transformed graph')
+      pkl_fname = "{}_transformed.pkl".format(ugraph.name)
+      with open(pkl_fname, 'wb') as fid:
+        pickle.dump(new_ugraph, fid)
+      logger.info('{} saved'.format(pkl_fname))
+    return new_ugraph
+  
+  @classmethod
+  def _check_non_quantized(cls, ugraph):
```
mbartling, line 38: This method should check if input or output tensors have quantization params. It probably should not be tied to operator names

dboyliao: Is it reasonable that we consider the model is quantized if there is any tensors in the model has quantization parameters?

utensor_cgen/legalizer/tflite.py
```diff
@@ -0,0 +1,138 @@
...
+  def legalize_ops(self, ugraph):
+    _GraphRewrite.apply(ugraph)
+    _OpTypeRename.apply(ugraph)
+    return ugraph
+  
+  def legalize_dtype(self, ugraph):
+    return ugraph
+
+
+class _OpTypeRename(object):
```
mbartling, line 25: These op names should not auto rename to quantized forms. This should be handled by the graph lowering stage.
`FullyConnected` -> `FullyConnectedOperator` as the generic form. Then the later stages can check quantization based on the tensors + quantization attributes.

dboyliao: I agree.
I'll see how to fix it.

dboyliao: But I don't think we should force the generic form to be unquantized.
Many models are given quantized and there is no guarantee we can convert quantized model back to a generic unquantized form.
I believe there is use case where the generic form is a quantized graph.

mbartling: I'm not suggesting forcing the generic form to be unquantized, I'm suggesting we leave the name to represent the op type and inject any attributes that can be used by the transform tools. For example:

```
Conv2D(input: float, filter: quantized int8, bias: int32)
```
Is probably not legal from a uTensor perspective so an error should be thrown since not all inputs are quantized.

But the following are both legal from a IR perspective, since they both represent the same functional operation even though their lowered targets are slightly different:
```
Conv2D(input: float, filter: float, bias: float); // Gets lowered to a Reference operator
Conv2D(input: quantized int8, filter: quantized int8, bias: int32) // Same op, but is_quantized is set to true so get's lowered to best form that accepts quantized inputs
```

Really the only stage between the legalizer and backend lowering is the Transformer pipeline, so as long as we capture intent of the op then the transforms can operate correctly. This way, the transforms can impose any additional constraints like all operators are quantized, or none, or a mix, without having to maintain some unique name for each type of op configuration.

dboyliao: I'm pretty sure I fixed the legalizer and lower.
Now the legalizer will not assuming a quantized graph and lower will change the op type once it find out the given graph is quantized.
BTW, one heuristic here is that I assume the graph is quantized if there is any tensor with quantization parameters.
I think this assumption is fair enough in most cases.

utensor_cgen/backend/transformer.py
dboyliao: I'll rename it to `_check_generic` and do nothing.
Once we have a full list of generic ops, I'll fill this method body.

utensor_cgen/legalizer/tflite.py
dboyliao: Update: the graph lower will quantize the graph op by op.

PR #109, opened 2020-05-12
[Do not merge] Re arch support extra ops

mbartling:
* Add additional operator support for target rearch.
* Fix bug in TFLITE parser where all tensors are treated as having quantization params.
* Add placeholder list for TF Lite side parser, so we can clarify and differentiate between frontend Op support and Backend Op support


mbartling:
> Of course, the conflicts need to be resolved too.

Done

PR #110, opened 2020-05-20
Tensor life cycle

dboyliao:
Add `Tensor::free` whenever appropriate.

### DIFF
--- a/utensor_cgen/backend/graph_lower/generic_graph_lower.py
+++ b/utensor_cgen/backend/graph_lower/generic_graph_lower.py
@@ -37,22 +37,22 @@ class TopoOrderTensorTimeslotPlanner(BackendPart):
     life_span = defaultdict(lambda: [None, None])
     for op_info in ugraph.ops_info.values():
       for tensor in op_info.input_tensors:
-        ref_cnts[tensor.name] += 1
+        ref_cnts[tensor] += 1
     for time_slot, op_name in enumerate(ugraph.topo_order):
       op_info = ugraph.ops_info[op_name]
       for tensor in op_info.output_tensors:
-        life_span[tensor.name][0] = time_slot
+        life_span[tensor][0] = time_slot
       for tensor in op_info.input_tensors:
-        ref_cnts[tensor.name] -= 1
-        if ref_cnts[tensor.name] == 0:
-          life_span[tensor.name][1] = time_slot
+        ref_cnts[tensor] -= 1
+        if ref_cnts[tensor] == 0:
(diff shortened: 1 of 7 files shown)


utensor_cgen/backend/graph_lower/generic_graph_lower.py
```diff
@@ -37,22 +37,22 @@ def apply(self, ugraph):
     life_span = defaultdict(lambda: [None, None])
     for op_info in ugraph.ops_info.values():
       for tensor in op_info.input_tensors:
-        ref_cnts[tensor.name] += 1
+        ref_cnts[tensor] += 1
```
mbartling, line 40: How do you handle inputs to the model? Pretty sure the logic will increment these but never decrement, but I just want to double check

utensor_cgen/backend/utensor/code_generator/rearch/_code_generator.py
```diff
@@ -80,7 +82,202 @@ def _find_required_ops(self, ugraph):
...
+      )
+    out_tensor_var_names = [
+      tensor_var_map[tensor.name] for tensor in chain(*[
+        ugraph.ops_info[op_name].output_tensors
+        for op_name in ugraph.output_nodes
+      ])
+    ]
+
+    # dict: tensor name -> TimeslotAllocation
+    tensor_timeslots = ugraph.attributes.get(TopoOrderTensorTimeslotPlanner.KWARGS_NAMESCOPE)
```
mbartling, line 108: Should there be a try catch around this? What if the ugraph doesnt contain a TimeslotPlanner?

dboyliao: That's why I use `get`
The `get` method will return a default value, which is `None` in this case, if the key is not in the dictionary.
If you look at the next line, I do have a check on the returned value and if there is no such plan available, it will roll back to naive code generation.

utensor_cgen/backend/graph_lower/generic_graph_lower.py
dboyliao: Oh, I didn't deal with inputs in this case.
Because the model inputs is declared by the users, the users should be responsible for its creation and destruction.
And I do decrement the ref cnt so I can find the life span of the tensors:
https://github.com/uTensor/utensor_cgen/pull/110/files/5d030fb3ea9ada6b73d16f7f2682c3a3c872fcd4#diff-ad6830d2f958b649ebbb340c55ab59e8R46

PR #112, opened 2020-06-03
update supported ops namespaces

dboyliao:


mbartling:
I like the graph lowering part since I can see a clear way that we can improve that in the future with some strategy setup. But I'm not a huge fan of adding extra boiler plate to every single Op declaration and Snippet declarations. Would probably be easier if we have the following:

```python
class UsingOpSnippet(_SnippetBase):
  __template_name__ = 'snippets/rearch/declare_using_op.cpp'

  def __init__(self, op, nested_namespaces=None):
    _SnippetBase.__init__(self)
    if nested_namespaces is None:
      nested_namespaces = []
    else:
      nested_namespaces = list(nested_namespaces)
    op_type = op.op_type
    if nested_namespaces:
      op_type = "::".join(nested_namespaces + [op_type])
    self.template_vars['op_type'] = op_type
```

And then add a default `get_using_namespace snippet(...)` in the `Operator` base class which can throw a warning if namespaces are empty. 

Then the output could be something like this:

```c++
using namespace uTensor;
using ReferenceOperators::Conv2dOperator;
using ReferenceOperators::ReLUOperator;
using ReferenceOperators::ReshapeOperator;
using ReferenceOperators::Conv2dOperator;
using ReferenceOperators::MaxPoolOperator;

void compute_model_v3(Tensor& t_conv2d_1_input0, Tensor& t_dense_1Sigmoid0){
    // start rendering local snippets
    ReferenceOperators::Conv2dOperator<float> op_000({ 1, 1, 1, 1 }, VALID);
    ReferenceOperators::ReLUOperator<float> op_001;
    ReferenceOperators::ReshapeOperator<float> op_003({ 1, 392 });
    ReferenceOperators::Conv2dOperator<float> op_005({ 1, 1, 1, 1 }, SAME);
    ReferenceOperators::MaxPoolOperator<float> op_006({ 2, 2 }, { 1, 2, 2, 1 }, VALID);
```

What do you think?

dboyliao:
> I like the graph lowering part since I can see a clear way that we can improve that in the future with some strategy setup. But I'm not a huge fan of adding extra boiler plate to every single Op declaration and Snippet declarations. Would probably be easier if we have the following:
> 
> ```python
> class UsingOpSnippet(_SnippetBase):
>   __template_name__ = 'snippets/rearch/declare_using_op.cpp'
> 
>   def __init__(self, op, nested_namespaces=None):
>     _SnippetBase.__init__(self)
>     if nested_namespaces is None:
>       nested_namespaces = []
>     else:
>       nested_namespaces = list(nested_namespaces)
>     op_type = op.op_type
>     if nested_namespaces:
>       op_type = "::".join(nested_namespaces + [op_type])
>     self.template_vars['op_type'] = op_type
> ```
> 
> And then add a default `get_using_namespace snippet(...)` in the `Operator` base class which can throw a warning if namespaces are empty.
> 
> Then the output could be something like this:
> 
> ```c++
> using namespace uTensor;
> using ReferenceOperators::Conv2dOperator;
> using ReferenceOperators::ReLUOperator;
> using ReferenceOperators::ReshapeOperator;
> using ReferenceOperators::Conv2dOperator;
> using ReferenceOperators::MaxPoolOperator;
> 
> void compute_model_v3(Tensor& t_conv2d_1_input0, Tensor& t_dense_1Sigmoid0){
>     // start rendering local snippets
>     ReferenceOperators::Conv2dOperator<float> op_000({ 1, 1, 1, 1 }, VALID);
>     ReferenceOperators::ReLUOperator<float> op_001;
>     ReferenceOperators::ReshapeOperator<float> op_003({ 1, 392 });
>     ReferenceOperators::Conv2dOperator<float> op_005({ 1, 1, 1, 1 }, SAME);
>     ReferenceOperators::MaxPoolOperator<float> op_006({ 2, 2 }, { 1, 2, 2, 1 }, VALID);
> ```
> 
> What do you think?

My design is trying to follow the single responsibility principle:
1. `Operator` only take care about the template signature. I implement singleton pattern for `Operator` so if the type signatures are the same, the same `Operator` instance will be returned.
2. `Snippet` only take care how the generate snippet given the parameters. It will not make any assumptions on what should or should not be given.

As a result, we can be sure the true logic will lie in only the code generator which using these objects as building blocks.

As for the using namespaces snippet method, I think it's much clear the namespaces are declared closely to the operator.
And by doing so, it will be clear all these ops are actually targeting different runtime operators under different namespaces.
In my opinion, this make it clear which operator should be mapped to specific runtime operator instead of a method call and check or modify it in the function body.

However, the down side, I think we will see many repeated pattern in the future since our runtime is going to do the best effort to keep parameters ordering, method signatures ...etc to be aligned across namespaces.
That may be a good sign to refactor our code as you suggest.
But such alignment is not a guarantee since we all agree it's going to be a soft requirement.
Therefore I prefer the keep the code simple and stupid at least for now.

PR #122, opened 2020-09-18
Fix pb file

dboyliao:
This is a fix for backward compatibility of `pb` file.
Because of all the scheme/op name changes from TF 1.x to TF 2.x, there may be corner cases that are not included in the tests.
This fix is not guaranteed to work.

We'll deprecate pb file in near future in favor of TF 2.0


dboyliao:
@mbartling
If you won't mind, I'm about to merge this PR.

PR #125, opened 2020-10-13
Fix tensor name

### LINKED ISSUE #124: Naming issue in utensor_cgen
The generated model code had around 51 errors where instead of writing for example t_sequential_1dense_9BiasAddReadVariableOpresource0 it would write t_sequential_1dense_9;BiasAddReadVariableOpresource0 so I had to remove the ; in each Tensor name.

dboyliao:
Fix #124

### DIFF
--- a/utensor_cgen/frontend/tflite.py
+++ b/utensor_cgen/frontend/tflite.py
@@ -79,7 +79,7 @@ class TFLiteParser(Parser):
 
     for idx in range(0, subgraph.TensorsLength()):
       tensor = subgraph.Tensors(idx)
-      tensor_name = tensor.Name().decode('utf8')
+      tensor_name = tensor.Name().decode('utf8').replace(';', '__')
       if tensor_name is "" or None:
         tensor_name = "tensor_" + str(idx)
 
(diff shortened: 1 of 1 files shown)


utensor_cgen/frontend/tflite.py
```diff
@@ -79,7 +79,7 @@ def _build_tensor_map(self, fb_model, ugraph):
 
     for idx in range(0, subgraph.TensorsLength()):
       tensor = subgraph.Tensors(idx)
-      tensor_name = tensor.Name().decode('utf8')
+      tensor_name = tensor.Name().decode('utf8').replace(';', '__')
```
mbartling, line 82: Do you know if there are any other possible corner cases? It might help to display a warning here if we run into an odd literal in parsing the tensor names

mbartling (approved):

utensor_cgen/frontend/tflite.py
mbartling: @dboyliao bump

dboyliao: Honestly, I don't know.
I really don't want to dig deep into tflite converter's C code for the tensor name issue here.
What kind of warning?
like this?
```python
if ';' in tensor.Name():
    # warning code here
    .... do stuff
```

mbartling: Yeah, just something that warns “Unexpected character, %s, found in tensor name”

dboyliao: Ok, done.
Screen Shot 2020-10-28 at 10 07 50 PM

I'll merge this PR, ASAP

Issue #127, opened 2020-11-18
Feature: add TransposeOperator


mbartling:
^^^

dboyliao:
Solved by #129

PR #129, opened 2020-11-20
Transpose op

dboyliao:
Solve #127

Using `TransposeOperator` in `uTensor` [PR 218](https://github.com/uTensor/uTensor/pull/218)


dboyliao:
Merge.
Close #127

PR #130, opened 2020-11-24
hotfix: reshape op data parser, inferring new shape from input tensors

dboyliao:
new shape sometime is input tensor in TFLite model file.
That's why the tflite parser will fail randomly since the op options will be `None` in this case.

Screen Shot 2020-11-24 at 4 52 46 PM


victorromeo:
Manually implementing this hot fix then running the unit tests in #133 shows this fix corrects the breaking test.
See `test_api/test_keras_tflite/test_export.py test_keras_model`

PR #133, opened 2020-11-30
Unit tests for ONNX export/convert

victorromeo:
Example of breaking tests


Pipfile
```diff
@@ -4,20 +4,162 @@ verify_ssl = false
 name = "pypi"
 
 [packages]
-e1839a8 = {path = ".",editable = true}
+absl-py = "==0.9.0"
```
dboyliao, line 7: Can you make the `packages` section clean as before?
As far as I know, the original pattern is better in package development.

dboyliao: If there is any testing package required for this, please add it in `setup.py`.
Or, move them to `dev-packages` section, via `pipenv install -d <package>`

utensor_cgen/api/export.py
```diff
@@ -42,3 +49,77 @@ def tflm_keras_export(
...
+          model_or_path
+        )
+      )
+
+    onnx_path = os.path.join(dir_path,f"{model_name}.onnx")
+    
+    torch.onnx.export(model, 
+      representive_dataset, 
+      onnx_path, 
+      verbose=True
```
dboyliao, line 80: I suggest to make `verbose` an argument so the caller can have control over the verbosity of this export api.

utensor_cgen/api/export.py
```diff
@@ -42,3 +49,77 @@ def tflm_keras_export(
...
+    )
+
+    # Print a human readable representation of the graph
+    onnx_model = onnx.load(onnx_path)
+    onnx.checker.check_model(onnx_model)
```
dboyliao, line 85: Maybe it's also better if the export api only print human readable representation if the `verbose` argument is `True`.

utensor_cgen/api/export.py
```diff
@@ -42,3 +49,77 @@ def tflm_keras_export(
...
+          model_or_path
+        )
+      )
+
+    # Perform Keras to ONNX conversion
+    onnx_model_name = f'{model_name}.onnx'
+    onnx_model = keras2onnx.convert_keras(model, model.name)
+    onnx.save_model(onnx_model, onnx_model_name)
+
+    # with (dir_path / 'tflm_model.tflite').open('wb') as fid:
```
dboyliao, line 122: remove these comments plz.

dboyliao (changes requested):
Overall looks good to me.
Minor request for changes required.

victorromeo:
Thanks for your feedback.  I really appreciate it.  To confirm these tests are failing at the moment, as ONNX doesn't contain the shape information required to support the uTensor_cgen operations using the ONNX frontend, so I think these tests will require further work.

dboyliao:
@victorromeo That's ok.
With your tests, I'll take a look this weekend to see if I can fix it.
BTW, may I ask why you or your team want to avoid tflite?
Just out of curiosity. XD

victorromeo:
The main reason is to remain as agnostic as possible if the researcher or developer intends to use a framework of choice.  The use of ONNX in utensor is a great starting point to achieve this.  We have found that different NN frameworks (TensorFlow, PyTorch, MXNet, etc) have resulted in different accuracy outcomes, due to minor implementation eccentricities.  As a side benefit, TF appears to undergo more rapid change, and therefore tools such as utensor are more fragile when they rely upon it.

dboyliao:
I agree.
TF has been the main root cause to issues in `utensor_cgen` for a long time.
Not to mention things get worse if you need to trace their python/c++ codes.
The code for eager mode is of all evils =   =a

utensor_cgen/frontend/onnx.py
```diff
@@ -20,13 +20,16 @@
...
 def _convert_op_attribute(attrib_pb):
+  ints_attribs = ['perm','dilations','kernel_shape','strides','pads']
   # TODO: integrate with ir.converter
   if attrib_pb.HasField('f'):
     return attrib_pb.f
   elif attrib_pb.HasField('i'):
     return attrib_pb.i
   elif attrib_pb.HasField('s'):
     return attrib_pb.s
+  elif attrib_pb.name in ints_attribs:
```
dboyliao, line 31: After double-check the `onnx` [proto3 file](https://github.com/onnx/onnx/blob/master/onnx/onnx-ml.proto#L166-L168), I suggest you get the `ints` attribute via `elif attrib_pb.HasFields('ints'): ....` which is more robust than listing.

dboyliao (changes requested):

utensor_cgen/frontend/onnx.py
dboyliao: Which will be better if you add `floats` for me XD

victorromeo: Added, however HasFields only works with scalar attributes, so a more indirect path was required.

dboyliao: You are right.
I don't know `HasField` is not working on repeated field (why protobuf doing this???)

I think maybe it's better to parse it by the `type` attribute:
https://github.com/onnx/onnx/blob/master/onnx/onnx-ml.proto#L154

That is:
```python
....
elif attrib_pb.type == 7:
    # ints
elif attrib_pb.type == 6:
    # floats
....
```

How do you think?

utensor_cgen/backend/utensor/code_generator/rearch/_operators/_impls.py
```diff
@@ -799,5 +836,5 @@ def get_eval_snippet(self, op_var_name, op_info, tensor_var_map):
   def get_construct_snippet(self, op_var_name):
     return None
 
-
-OperatorFactory._operators[_MissingOperator.op_type] = _MissingOperator
+for namespace in _MissingOperator.namespaces:
```
dboyliao, line 839: Why do you need to add these loop?
It will break the code generator.

dboyliao (changes requested):

utensor_cgen/backend/utensor/code_generator/rearch/_operators/_impls.py
dboyliao: Oh, I think of a better way to do this.
Please revert this loop.
I'll make update to `develop` branch.

victorromeo: Added #136 to track the fix

victorromeo: Reverted

utensor_cgen/frontend/onnx.py
victorromeo: Minor cleanup done as follows, as this checks member existence and len > 1:
....
elif attrib_pb.floats:
    # floats
elif attrib_pb.ints:
    # ints
elif attrib_pb.strings:
    # strings
....

victorromeo (commented):
Reverted

dboyliao:
CI blocks....
I've run the tests manually but it fails.
Since it's mostly due to frontend parser, the export api itself should be good.
I'll merge it and work on fixing frontend afterward.

PR #138, opened 2020-12-01
Remove legacy

dboyliao:
remove legacy
    - update matcher
        - matcher api now works on rearch ops
    - update graph builder
        - graph builder will build generic graph with op types as rearch ops
    - update tests
    - remove related python modules
        - legacy backend code generator is GONE!

As a side note, the builder api still using tensorflow 1.x api to do some tensor shape/dtype inference.
Not hard to fix but I just want to keep things simple in this PR.
We'll transit to tensorflow 2.x.

Eventually :p


tests/test_api/test_keras_onnx/test_export.py
```diff
@@ -3,9 +3,11 @@
 from pathlib import Path
 
 import pytest
+
 import utensor_cgen.api.export as export
-import tensorflow.keras as keras
 
+
+@pytest.mark.beta_release
```
victorromeo, line 10: Great option to add this. Nice work.

utensor_cgen/transformer/quantize.py
```diff
@@ -1,37 +0,0 @@
-"""Legacy, DON'T USE
```
victorromeo, line 1: May I please confirm the intent behind deprecating the quantization transformation?  Is it intended that the user will provide a pre-quantized model?  Other than tflite int8 quantization, which external quantization mechanisms are recommended?

utensor_cgen/legalizer/tflite.py
```diff
@@ -7,6 +7,7 @@
 
 from .api import Legalizer
 from .base import LegalizerBase
+from .utils import _hotfix_reshape
```
victorromeo, line 10: May be worth renaming this operation at some stage

victorromeo (commented):
Seems to have a bit of dead code behind comments still. Recommend general once over.

utensor_cgen/api/export.py
```diff
@@ -1,15 +1,12 @@
+import os
 import tempfile
 from pathlib import Path
 
-import os
-
+import keras2onnx
```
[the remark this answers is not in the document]
dboyliao: You mean lazily import `keras2onnx`?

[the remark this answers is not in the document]
victorromeo: Yes. Your use of the LazyLoader is pretty good also.  As these packages are only used by specific export operations, it makes sense to keep using that pattern.

utensor_cgen/transformer/quantize.py
dboyliao: Yes, it's intended that user can have option to pass quantized or pre-quantized model.
And yes, I simply forget to remove this one in this PR.

We are not deprecate quantization but this transformer is almost of no use in TF 2.x
We have some breaking changes, due to TF 2.x, to the `uTensorGraph` so the conversion to `tf.GraphDef` is broken.
Such conversion is required for this transformer and the `TransformGraph` in TF is removed in TF 2.

victorromeo: Ok. Thanks for confirming

utensor_cgen/legalizer/tflite.py
dboyliao: Ya. We are not sure if this hotfix is a good idea or not.
Let's leave it as is for now.

utensor_cgen/api/export.py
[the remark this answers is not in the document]
dboyliao: I only use it when the submodule is used in cli, where some import may slow down cli loading time and bring bad user experience, such as waiting for long time for help messages.
Native import will help development in ways of editor which can leverage import statements to find definitions in the program.
As a result, I prefer to use native import statements here.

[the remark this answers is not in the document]
dboyliao: Oh, btw, I will also use lazy import when there is cyclic import in python, which is quite common for a project like this one.

PR #139, opened 2020-12-07
MaxPool op in list-support-ops

victorromeo:
Noticed that the MaxPool operator wasn't listed when using list-support-ops.


dboyliao:
Oh, it was my mistake.
You should change `_GenericMaxOperator` to `_GenericMaxPoolOperator`.
That should fix it.

dboyliao:
BTW, these generic operators are defined in `_graph_builder.py` which is related to building graph in `utensor_cgen`.
The modules are arranged in this way since I'd like to keep graph building api to be generic.
That is, the graph built with `utensor_cgen` should be generic and can be lowered to other target by users.

dboyliao:
@victorromeo, thanks for reporting the issue and the PR.
However, there is better way to fix the issue other than this PR and it's fixed at fa8cdb4882.
Closing this PR.

